PR #5, opened 2022-03-15
corrections to hexl code discovered 2022-03-15

jbates-duality:
First round of HEXL mods.  This allows core and its unit tests to build successfully.


dsuponitskiy (approved):
Looks good

PR #12, opened 2022-03-25
Sphinx logging best practices update

iquah1:
Turns out the previous push only stopped Sphinx from spitting out errors during a local build. On RTD it seems that they call Sphinx on their own. This PR aims to fix that too


iquah1:
This is the PR that I mentioned earlier where I'd like to try and get this merged in today

.readthedocs.yaml
```diff
@@ -0,0 +1,26 @@
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.6"
+    # You can also specify other tool versions:
```
mtriplett-duality, line 9: For other tool versions, do we want to add things like doxygen, graphviz, gcc/g++? I'm not familiar with the purpose of this particular .yaml file so I'm not sure what tools/toolchain it's specifically referring to. Is it just for the sphinx work or the whole build+doc build?

mtriplett-duality (approved):

mtriplett-duality:
Additions all look good, might need some CI tweaking but the branch itself looks good.

.readthedocs.yaml
iquah1: Those are really good questions, and unfortunately I'm not sure. I don't know if we need those things and unfortunately I don't think we'll know until we test it out. I've tried looking for a way to test it without first merging the PR but I haven't found a way

PR #16, opened 2022-03-28
Add docs option to CMake 

mtriplett-duality:
Add CMake option for docs target, so docs aren't added to the main lib target unless the option is set.


mtriplett-duality:
Before merging, must add BUILD_DOCS option for pages job in CI

iquah1 (approved):

PR #22, opened 2022-04-03
CI bug fix

dsuponitskiy:


jbates-duality (approved):
Dmitriy assures me that this near-trivial change compiles and runs.

CI checks pass.

I am now approving.

PR #24, opened 2022-04-05
Changes for clang-format and cpplint

dsuponitskiy:


jbates-duality:
I've randomly sampled the changes to several files in binfhe and benchmark.

I agree with 99% of the reformatting.

I also looked at the "remove using namespace std" patch.  This appears cleanly-done.

I see that there are some conflicting files.  I have not gone down that rabbithole.  It's probably best for Dmitriy to do this.

jbates-duality:
I have just reviewed the conflicting files.

It would appear that these are related to Ian's documentation project, and should be simple merges.

jbates-duality (approved):
Approving this now that conflicts are resolved and CI passed.

benchmark/src/compare-bfvrns-vs-bgvrns.cpp
```diff
@@ -60,119 +59,119 @@ usint mult_depth = 3;
...
 
-  std::vector<int64_t> vectorOfInts = {1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
-  Plaintext plaintext = cc->MakeCoefPackedPlaintext(vectorOfInts);
+    std::vector<int64_t> vectorOfInts = {1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+    Plaintext plaintext               = cc->MakeCoefPackedPlaintext(vectorOfInts);
 
-  vector<Ciphertext<DCRTPoly>> ciphertexts;
-  for (int i = 0; i < (1 << mult_depth); i++)
-    ciphertexts.push_back(cc->Encrypt(keyPair.publicKey, plaintext));
+    vector<Ciphertext<DCRTPoly>> ciphertexts;
```
mtriplett-duality, line 125: why doesn't this require "std::"? I thought we removed the namespace up top

mtriplett-duality (approved):

dsuponitskiy:
There is more cleanup for the "std::" stuff , but I will fix later. This branch initially was for reformatting only.

benchmark/src/compare-bfvrns-vs-bgvrns.cpp
jbates-duality: aye, yaye, yaye.  Good catch.  Regex: [^:]vector has 677 hits in the codebase.

PR #27, opened 2022-04-09
Removal of "using namespace std;", "using std::vector;", using std::string;", etc.

dsuponitskiy:


jbates-duality (approved):
You are a machine.  The renaming of symbols that are keywords will undoubtedly help someone's sanity.  This [large] patch is straightforward, and as-advertised.  I pulled and built it locally.  CI passed.  I approve.

dsuponitskiy:
Thank you

PR #32, opened 2022-04-13
Fix uninitialized memory issues

ericastor:
m_timeOptimization is not initialized on all paths, so risks uninitialized memory problems unless initialized by default.

### DIFF
--- a/src/binfhe/include/binfhecontext.h
+++ b/src/binfhe/include/binfhecontext.h
@@ -411,7 +411,7 @@ private:
     std::map<uint32_t, RingGSWEvalKey> m_BTKey_map;
 
     // Whether to optimize time for sign eval
-    bool m_timeOptimization;
+    bool m_timeOptimization = false;
 };
 
 }  // namespace lbcrypto


iquah1:
@ericastor 

Would you say this PR is done and has addressed the general issue of uninitialized memory? I assume you've tested it and it has

1) produced the behavior you'd like
2) not broken any existing tests

? Asking because you haven't tagged anyone for review

ericastor:
Who should I generally tag for review on this sort of PR? Yes, I'd consider it done.

iquah1:
Good question! 

> Who should I generally tag for review on this sort of PR? Yes, I'd consider it done.

I'd say @mtriplett-duality for now. 

Also, do review our [contributing guidelines](https://openfhe-development.readthedocs.io/en/latest/assets/sphinx_rsts/contributing/rebasing_off_main.html) for future PRs (we're updating our documentation as we move along so not all of it will be relevant)

ericastor:
Sure! I don't believe I have write access to the repository - so it looks like I can't tag anyone specific in for review on a PR, or work on a feature branch within this repo.

If we should change that, let me know!

iquah1:
Gotcha! Thanks, I pinged the internal team and we'll get a response soon I think

src/binfhe/include/binfhecontext.h
```diff
@@ -411,7 +411,7 @@ class BinFHEContext : public Serializable {
     std::map<uint32_t, RingGSWEvalKey> m_BTKey_map;
 
     // Whether to optimize time for sign eval
-    bool m_timeOptimization;
+    bool m_timeOptimization = false;
```
dsuponitskiy, line 414: This data member should be initialized in constructors.

dsuponitskiy: Sorry, I may be wrong as  C++ may allow initialization of non-static or not-const data members this way. I need to double check this

dsuponitskiy: Yes, I am wrong. It is allowed

iquah1: If you're happy with it as is would you mind merging it? It sounds like you are

yspolyakov (approved):

PR #37, opened 2022-04-25
Unittestfix

kimandrik:
fixed several tests, including compress, reEncryption and workaround for ciphertext/plaintext operations


dsuponitskiy (approved):
Looks good

PR #40, opened 2022-05-03
Bug fix for key switching; changes for unit tests

dsuponitskiy:


dsuponitskiy:
Added an exception to cryptocontext.cpp. For 3 other files: the logic stays as an exception may not be thrown from there.

jbates-duality (commented):
Looks good to me.  I'll approve and merge when CI completes.

jbates-duality (approved):
CI passed.  Approving and merging now.

PR #44, opened 2022-05-06
Set standard deviation's default value to 3.19

dsuponitskiy:
1. Set standard deviation's default value to 3.19 in src/pke/include/scheme/cryptocontextparams-defaults.h
2. Remove calls to SetStandardDeviation() setting the value to 3.19 or 3.2
3. Replace all values of 3.19 or 3.2 with DFLT in unit tests


jbates-duality (approved):
The intent of this patch is clear: standardize a default sigma / std. dev. amongst the various scnemes.  I do not see any issue with it.  I am approving.

PR #49, opened 2022-05-12
Removal of default parameters in virtual functions

dsuponitskiy:


jbates-duality (approved):
The intent of this patch is clear and threefold:

1) remove default parameters from virtual methods.
2) patch calling code to not depend upon default parameters that have been removed.
3) change certain methods from virtual to non-virtual

I have inspected the code and found it good.  I have pulled the tip of this branch and performed a default build, with no warnings or errors.  The three binaries in unittest ran successfully.

I am approving this patch.

jbates-duality:
After consultation with DS, I am squashmerging this.

PR #62, opened 2022-06-05
Throw an exception if plaintextModulus for BGV is zero

dsuponitskiy:


jbates-duality (approved):
Well, this is straightforward.

Approving.

PR #65, opened 2022-06-08
fixed a bug with bgv-flexautoext

yspolyakov:


yspolyakov:
@sarojaduality I think we should run the for loop to compute the scaling factor in BGV *Decrypt* only for FLEXIBLEAUTO*. Currently it does it for all modes.

dsuponitskiy (approved):
Addressed comments

PR #77, opened 2022-06-29
Corrected comparison of the approximation errors between FLEXIBLEAUTO…

dsuponitskiy:
… and FLEXIBLEAUTOEXT


jbates-duality (approved):
This seems pretty straightforward.  Changing error approxmation calculation to floating point, and bumping up the ring dimension in unit test scenarios.  I'm approving now.

PR #79, opened 2022-06-30
Fixed BEHZ bug.

sarojaduality:


src/core/include/lattice/hal/dcrtpoly-interface.h
```diff
@@ -1258,6 +1259,7 @@ class DCRTPolyInterface : public ILElement<DerivedType, BigVecType> {
    * @param &BModqPrecon NTL precomptations for [B]_{q_i}
    */
     virtual void FastBaseConvSK(
+        const std::shared_ptr<Params> paramsQ,
         const std::vector<NativeInteger>& moduliQ, const std::vector<DoubleNativeInt>& modqBarrettMu,
```
yspolyakov, line 1263: @sarojaduality Do we still moduliQ? We should be able to read them from paramsQ.

yspolyakov (approved):

src/core/include/lattice/hal/dcrtpoly-interface.h
sarojaduality: Removed

PR #82, opened 2022-06-30
nPrimes removal

dsuponitskiy:


jbates-duality (approved):
This seems straighforward, removing nPrimes class members, and the related fanout.  Approving.

PR #83, opened 2022-07-01
assuranceMeasure and scaleExp removal

dsuponitskiy:


jbates-duality (approved):
Patch is removing assuranceMeasure from params in favor of member variables.  Straightforward.  Approving.

PR #90, opened 2022-07-06
Initial CKKS bootstrapping implementation

yspolyakov:


src/pke/include/schemebase/base-fhe.h
```diff
@@ -0,0 +1,111 @@
...
+   *
+   * @param levelBudget - vector of budgets for the amount of levels in encoding
+   * and decoding
+   * @param dim1 - vector of inner dimension in the baby-step giant-step routine
+   * for encoding and decoding
+   * @param slots - number of slots to be bootstrapped
+   * @param debugFlag - set to 1 when debugging encoding/decoding only
+   * @param precomp - do linear transform precomputations
+   */
+  virtual void EvalBootstrapSetup(
```
yspolyakov, line 74: We should move the default values to the CKKS-specific implementation

PR #92, opened 2022-07-07
BGV moduli selection for other modes

sarojaduality:


src/pke/include/scheme/bgvrns/bgvrns-parametergeneration.h
```diff
@@ -110,8 +110,24 @@ class ParameterGenerationBGVRNS : public ParameterGenerationRNS {
                      usint digitSize,
                      uint32_t auxBits,
                      enum KeySwitchTechnique ksTech,
+                     enum RescalingTechnique rsTech,
                      usint numPrimes) const;
 
+  std::pair<std::vector<NativeInteger>, uint32_t> computeAutoModuli(std::shared_ptr<CryptoParametersBase<DCRTPoly>> cryptoParams,
```
yspolyakov, line 116: As far as I can see, computeAutoModuli and computeFlexibleAutoModuli have a lot in common. Probably better use a single method for them.

sarojaduality: Combined them into a single method, and got rid of some of the extraneous parameters.

yspolyakov (approved):

PR #97, opened 2022-07-12
updated broken URLS for static docs, and removed doc build off cmake

iquah1:
Changes:

- [x] modified the `make uninstall` command which was pointing to the wrong place
- [x] removed docs from build process
- [x] Updated URLs for static docs


iquah1:
@yspolyakov 

Looks good to you?

PR #116, opened 2022-07-16
Cleanup of PKE examples and user API

yspolyakov:
* Adds EvalRotate (currently calls EvalAtIndex)
* Updates openfhe.h so that only one header would need to be included by the user
* Updates PKE examples and extras


dsuponitskiy (approved):
Looks good

PR #118, opened 2022-07-17
Documentation updates for the initial release

yspolyakov:


docs/sphinx_rsts/intro/using_the_library.rst
```diff
@@ -7,33 +7,19 @@ The source files in the demo directories, particularly those in src/examples/pke
...
-
   * ``BFVrns``
 
-  * ``BFVrnsB``
-
   * ``BGVrns``
 
-  * ``CKKS``
-
-  * ``Null`` (a test scheme which does not actually encrypt/decrypt anything)
```
iquah1, line 30: Did we remove the `Null` scheme? Is there a reason why we did it?

iquah1 (commented):
Just one question but looks good to me. Let's merge after looking at the output and verifying everything, but so far I don't think that it will break any formatting

docs/sphinx_rsts/intro/using_the_library.rst
yspolyakov: yes, we removed it. The previous null scheme was only for BGV and BFV. It was more confusing to keep it.

iquah1 (approved):
LGTM

PR #132, opened 2022-07-19
Fixed a clang complaint for 128-bit link on MacOS

dsuponitskiy:


jbates-duality (approved):
Straighforward change to appease clang.  Approving.

PR #142, opened 2022-07-21
operations with plaintexts without decoding and encoding again

### LINKED ISSUE #59: Operations Ciphertext +/-/* Plaintext without decoding the Plaintext
In LeveledSHECKKSRNS class several operations: ciphertext +/-/* plaintext are implemented using workaround.
First we decode the plaintext and then encode again with the proper depth/level. 

Instead, we can treat a plaintext as a "dummy" ciphertext, and do all operations in a same way as ciphertext * ciphertext.

P.S. we also can redefine the Plaintext to smth like PlainCiphertext. And define a Plaintext to be not depend on depth/level, and instead Plaintext will be 
- NativePoly (mod t) = InvNTT(integer message vector (mod t) )  for BGV/BFV 
- vector<double> = InvFFTSpecial(double message vector) for CKKS

kimandrik:
closes #59 
Added MorphPlaintext method that creates a ciphertext with a single element from a plaintext.
In UnitTests changed MakeCKKSPackedPlaintext depth from 3 to 2 (to be compatible with AUTO logic, where depth <= 2)
In UnitTests changed eps -> epsHigh where FIXEDAUTO mode precision become worse.

### DIFF
--- a/src/pke/lib/schemerns/rns-leveledshe.cpp
+++ b/src/pke/lib/schemerns/rns-leveledshe.cpp
@@ -125,11 +125,16 @@ void LeveledSHERNS::EvalAddInPlace(Ciphertext<DCRTPoly>& ciphertext, ConstPlaint
         return;
     }
 
-    DCRTPoly pt = cryptoParams->GetScalingTechnique() == FIXEDMANUAL ?
-                      AdjustLevelsInPlace(ciphertext, plaintext) :
-                      AdjustLevelsAndDepthInPlace(ciphertext, plaintext);
+    auto ctmorphed = MorphPlaintext(plaintext, ciphertext);
 
-    EvalAddCoreInPlace(ciphertext, pt);
+    if (cryptoParams->GetScalingTechnique() == FIXEDMANUAL) {
+        AdjustLevelsInPlace(ciphertext, ctmorphed);
+    }
+    else {
+        AdjustLevelsAndDepthInPlace(ciphertext, ctmorphed);
+    }
+
+    EvalAddCoreInPlace(ciphertext, ctmorphed->GetElements()[0]);
 }
 
 Ciphertext<DCRTPoly> LeveledSHERNS::EvalAddMutable(Ciphertext<DCRTPoly>& ciphertext, Plaintext plaintext) const {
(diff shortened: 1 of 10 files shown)


yspolyakov (approved):
Looks good to me

yspolyakov:
@dsuponitskiy please review the code. I've checked the logic.

src/pke/lib/schemerns/rns-leveledshe.cpp
```diff
@@ -125,11 +125,16 @@ void LeveledSHERNS::EvalAddInPlace(Ciphertext<DCRTPoly>& ciphertext, ConstPlaint
         return;
     }
 
-    DCRTPoly pt = cryptoParams->GetScalingTechnique() == FIXEDMANUAL ?
-                      AdjustLevelsInPlace(ciphertext, plaintext) :
-                      AdjustLevelsAndDepthInPlace(ciphertext, plaintext);
+    auto ctmorphed = MorphPlaintext(plaintext, ciphertext);
 
-    EvalAddCoreInPlace(ciphertext, pt);
+    if (cryptoParams->GetScalingTechnique() == FIXEDMANUAL) {
```
dsuponitskiy, line 130: This block is used in too many places. It should be a separate function

src/pke/lib/schemerns/rns-leveledshe.cpp
```diff
@@ -393,24 +417,24 @@ void LeveledSHERNS::EvalMultInPlace(Ciphertext<DCRTPoly>& ciphertext, ConstPlain
...
-    EvalMultCoreInPlace(ciphertext, pt);
-    ciphertext->SetDepth(ciphertext->GetDepth() + 1);
-    ciphertext->SetScalingFactor(ciphertext->GetScalingFactor() * ciphertext->GetScalingFactor());
+    ciphertext->SetDepth(ciphertext->GetDepth() + ctmorphed->GetDepth());
+    ciphertext->SetScalingFactor(ciphertext->GetScalingFactor() * ctmorphed->GetScalingFactor());
     if (cryptoParams->GetScalingTechnique() == FLEXIBLEAUTO || cryptoParams->GetScalingTechnique() == FLEXIBLEAUTOEXT) {
         const auto plainMod = ciphertext->GetCryptoParameters()->GetPlaintextModulus();
-        ciphertext->SetScalingFactorInt(
-            ciphertext->GetScalingFactorInt().ModMul(ciphertext->GetScalingFactorInt(), plainMod));
+        ciphertext->SetScalingFactorInt(ciphertext->GetScalingFactorInt().ModMul(
```
dsuponitskiy, line 435: This line is hard to read. Should be split in to 2 lines

src/pke/lib/schemerns/rns-leveledshe.cpp
```diff
@@ -422,25 +446,25 @@ Ciphertext<DCRTPoly> LeveledSHERNS::EvalMultMutable(Ciphertext<DCRTPoly>& cipher
...
 
+    auto ctmorphed = MorphPlaintext(plaintext, ciphertext);
+
     if (cryptoParams->GetScalingTechnique() == FIXEDMANUAL) {
-        Ciphertext<DCRTPoly> c = ciphertext->Clone();
-        DCRTPoly pt            = AdjustLevelsInPlace(c, plaintext);
-        auto result            = EvalMultCore(c, pt);
-        result->SetScalingFactor(c->GetScalingFactor() * plaintext->GetScalingFactor());
-        result->SetDepth(c->GetDepth() + plaintext->GetDepth());
+        Ciphertext<DCRTPoly> ct = ciphertext->Clone();
```
dsuponitskiy, line 452: The function should be re-written as logic is the same for  FIXEDMANUAL and for the block after. we just need to have an auto reference to the correct ciphertext.

dsuponitskiy (approved):
Looks good

dsuponitskiy:
Started 4 manual workflows for this PR

PR #150, opened 2022-07-25
Throw exception if any 2 of multiplicativeDepth, evalAddCount and key…

dsuponitskiy:
…SwitchCount are non-zeros

Changes:
1. Added an exception for the above condition
2. Changed the type from int32_t to uint32_t for multiplicativeDepth, evalAddCount and keySwitchCount
3. Made some ParameterGenerationBGVRNS' methods private as they are not used outside of the class. they can be made public again later if necessary
4. Fixed unit tests that were added incorrectly and failed because of the new check.

manually started 4 CI jobs.


src/pke/include/scheme/bfvrns/bfvrns-parametergeneration.h
```diff
@@ -47,8 +47,9 @@ class ParameterGenerationBFVRNS : public ParameterGenerationRNS {
 public:
     virtual ~ParameterGenerationBFVRNS() {}
 
-    bool ParamsGenBFVRNS(std::shared_ptr<CryptoParametersBase<DCRTPoly>> cryptoParams, int32_t evalAddCount,
-                         int32_t multiplicativeDepth, int32_t keySwitchCount, size_t dcrBits, uint32_t n,
+    // TODO (dsuponit): should we use uint32_t for evalAddCount, multiplicativeDepth and keySwitchCount?
```
sarojaduality, line 50: Remove leftover comment.

src/pke/lib/scheme/bfvrns/bfvrns-parametergeneration.cpp
```diff
@@ -288,6 +288,16 @@ bool ParameterGenerationBFVRNS::ParamsGenBFVRNS(std::shared_ptr<CryptoParameters
             }
         }
     }
+    else if (multiplicativeDepth && (evalAddCount || keySwitchCount) ||
+             keySwitchCount && (multiplicativeDepth || evalAddCount)) {
```
sarojaduality, line 292: Should this be changed to keySwitchCount && evalAddCount?

sarojaduality (approved):

PR #162, opened 2022-08-09
Refactor class Serializable

### LINKED ISSUE #157: class Serializable refactoring
1. In Serializable: implement virtual SerializedObjectName() to return the correct type name of any object derived from Serializable and remove all other implementations of SerializedObjectName().
2. In Serializable: implement static SerializedVersion() and remove all other implementations of SerializedVersion()
3. Make sure that:
- all classes having save() and load() implemented are derived from Serializable 
- the version number is checked in every load() function (as some classes miss a. or b. or both of them)

dsuponitskiy:
Closes #157

### DIFF
--- a/src/binfhe/include/binfhecontext-ser.h
+++ b/src/binfhe/include/binfhecontext-ser.h
@@ -40,13 +40,13 @@
 #include "utils/serial.h"
 
 // Registers types needed for serialization
-CEREAL_REGISTER_TYPE(lbcrypto::LWECryptoParams);
-CEREAL_REGISTER_TYPE(lbcrypto::LWECiphertextImpl);
-CEREAL_REGISTER_TYPE(lbcrypto::LWEPrivateKeyImpl);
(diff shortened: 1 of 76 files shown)


jbates-duality:
I've run benchmarks on tip of dev, and tip of the branch for this PR.  I am observing significant performance degradation.  Unable to approve at this time.

jbates-duality:
As due-diligence here, I have compared the lib-benchmark and poly-benchmark-4k output for tip-of-dev and tip of this branch.  Here are my findings: Some NTTs in lib-benchmark have 3-5% performance degradation.  The deepest degradation is: NTTTransformInPlace, which is showing ~6% degradation.  After running this benchmark several times, I believe that this degradation is real, and not a mirage or noise.  poly-benchmark-4k shows no degradation, and possibly a very small performance improvement.

yspolyakov:
Requires a lot of work. Will be done after v1.0 is released.

PR #177, opened 2022-09-06
Fix issues 94 110

sararv22:
changed getparamspk() in rns-cryptoparameters.h to check for both keyswitching method hybrid and pre mode is set to fix issue 110.

changed numPrimes+1 to numPrimes in computeNoiseEstimates function in bgvrns-parametergeneration.cpp and verified that the multihopPRE unittest passes to fix issue 94


src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
```diff
@@ -110,7 +110,7 @@ BGVNoiseEstimates ParameterGenerationBGVRNS::computeNoiseEstimates(
         int relinBase       = pow(2.0, digitSize);
         int modSizeEstimate = DCRT_MODULUS::MAX_SIZE;
         int numWindows      = floor(modSizeEstimate / log(relinBase)) + 1;
-        keySwitchingNoise   = numWindows * (numPrimes + 1) * expansionFactor * relinBase * Berr / 2.0;
+        keySwitchingNoise   = numWindows * (numPrimes) * expansionFactor * relinBase * Berr / 2.0;
```
dsuponitskiy, line 113: numPrimes should not be wrapped in parenthesis.

src/pke/lib/scheme/bfvrns/bfvrns-parametergeneration.cpp
```diff
@@ -62,7 +62,7 @@ bool ParameterGenerationBFVRNS::ParamsGenBFVRNS(std::shared_ptr<CryptoParameters
     MultiplicationTechnique multTech = cryptoParamsBFVRNS->GetMultiplicationTechnique();
     ProxyReEncryptionMode PREMode    = cryptoParamsBFVRNS->GetPREMode();
 
-    if (PREMode != INDCPA) {
+    if ((PREMode != INDCPA) && (PREMode != NOT_SET)) {
         OPENFHE_THROW(not_available_error, "HRA modes for PRE with BFVRNS not implemented yet");
```
dsuponitskiy, line 66: i would throw an exception with a different error message which includes the value of PREMode. something like this:
"this ProxyReEncryptionMode " + PREMode + " is not supported for BFVRNS". The same should apply to
src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp: l.334
src/pke/lib/scheme/ckksrns/ckksrns-parametergeneration.cpp: l.62

src/pke/unittest/utils/UnitTestCCParams.h
```diff
@@ -63,6 +63,7 @@ struct UnitTestCCParams {
     double   keySwitchCount; // BFVRNS, 
     double   multiplicationTechnique; // BFVRNS, 
     double   encryptionTechnique; // BFVRNS,
+    double PREMode; // BGVRNS, BFVRNS, CKKSRNS
```
dsuponitskiy, line 66: is PREMode supposed to be tested for a limited number of unit tests? if yes, then PREMode should be added as an additional test case data to "struct TEST_CASE_UTxxxxxx". if PREMode may be added to any unit test, then ALL unit tests in ALL files should get the value of DFLT for PREMode.

dsuponitskiy (approved):
Looks good now

PR #186, opened 2022-09-20
Making schemeId enum

dsuponitskiy:
SchemeId should be enum, not std::string


jbates-duality (approved):
Straightforward replacement of string schemeId with an enum.  Approving.

PR #194, opened 2022-09-29
Resolve issue #99 (different number of towers in multiplication in CKKS bootstrapping)

### LINKED ISSUE #99: Mismatch in multiplying DCRTPolys
When the exception in DCRTPolyImpl<VecType>::Times is enabled, it gets triggered during CKKS bootstrapping. See example below. Need to fix the root cause and reenable the exception.

```
bin/examples/pke/ckks-bootstrap
SecretKeyDist: UNIFORM_TERNARY
p = 59
n = 4096
# slots = 2048
log2 q = 1948
log2 q*p = 2608
dim1 = [ 0 0 ]
level budget = [ 4 4 ]

Linear transform precomputation time: 1.72 s

Automorphism key generation time: 2.175 s
ciphertext number of slots: 2048

Number of levels before bootstrapping: 1
terminate called after throwing an instance of 'lbcrypto::math_error'
  what():  /home/ypolyakov/git/openfhe-development/src/core/lib/lattice/hal/default/dcrtpoly.cpp:914 tower size mismatch; cannot multiply
Aborted (core dumped)
```

sarojaduality:
Closes #99.

### DIFF
--- a/src/core/include/lattice/hal/dcrtpoly-interface.h
+++ b/src/core/include/lattice/hal/dcrtpoly-interface.h
@@ -584,6 +584,15 @@ public:
     virtual DerivedType Times(const std::vector<NativeInteger>& element) const = 0;
 
     /**
+   * @brief Performs a multiplication operation even when the multiplicands
+   * have a different number of towers.
+   *
+   * @param &element is the element to multiply with.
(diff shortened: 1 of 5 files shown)


dsuponitskiy (approved):
Looks good

PR #202, opened 2022-10-11
Improve CKKS bootstrapping precision for FLEXIBLEAUTOEXT.

### LINKED ISSUE #128: Improve CKKS bootstrapping precision for FLEXIBLEAUTOEXT
Currently the parameters for approximating the modular reduction using the sine wave in the 64-bit scenario are tailored for FLEXIBLEAUTO. A better precision can be achieved for FLEXIBLEAUTOEXT if a different (larger) ratio of q/m is used.

sarojaduality:
Closes #128


benchmark/src/ckks-bootstrapping-precision.cpp
```diff
@@ -0,0 +1,140 @@
+//==================================================================================
```
yspolyakov, line 1: It looks like an example rather than a Google benchmark file. I suggest moving this files to extras for pke.

src/pke/include/scheme/ckksrns/ckksrns-fhe.h
```diff
@@ -115,9 +115,10 @@ class CKKSBootstrapPrecom {
 class FHECKKSRNS : public FHERNS {
     using ParmType = typename DCRTPoly::Params;
 
-    const double K_SPARSE  = 14.0;   // upper bound for the number of overflows in the sparse secret case
-    const double K_UNIFORM = 512.0;  // upper bound for the number of overflows in the uniform secret case
-    static const int32_t R = 6;      // number of double-angle iterations in CKKS bootsrtrapping
+    const double K_SPARSE     = 14.0;   // upper bound for the number of overflows in the sparse secret case
+    const double K_UNIFORM    = 512.0;  // upper bound for the number of overflows in the uniform secret case
+    static const int32_t R    = 6;      // number of double-angle iterations in CKKS bootsrtrapping
+    double m_correctionFactor = 0.0;    // correction factor, which we scale the message by to improve precision
```
yspolyakov, line 121: why do we have a global variable here? this is not a constant.

src/pke/include/schemebase/base-fhe.h
```diff
@@ -74,9 +74,10 @@ class FHEBase {
    * @param dim1 - vector of inner dimension in the baby-step giant-step routine
    * for encoding and decoding
    * @param slots - number of slots to be bootstrapped
+   * @param correctionFactor - value to rescale message by to improve precision. If set to 0, we use the default logic.
```
yspolyakov, line 77: we should say this is only used for NATIVE_SIZE=64 (not used for NATIVE_SIZE=128)

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
```diff
@@ -61,6 +61,26 @@ void FHECKKSRNS::EvalBootstrapSetup(const CryptoContextImpl<DCRTPoly>& cc, std::
     uint32_t M     = cc.GetCyclotomicOrder();
     uint32_t slots = (numSlots == 0) ? M / 4 : numSlots;
 
+    // Set correction factor by default, if it is not already set.
+    if (correctionFactor == 0.0) {
+        if (cryptoParams->GetScalingTechnique() == FLEXIBLEAUTOEXT) {
```
yspolyakov, line 66: Please check if we can add FLEXIBLEAUTO here.

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
```diff
@@ -264,7 +284,7 @@ Ciphertext<DCRTPoly> FHECKKSRNS::EvalBootstrap(ConstCiphertext<DCRTPoly> ciphert
     double powP  = pow(2, p);
 
     double deg        = std::round(std::log2(qDouble / powP));
-    double correction = 9.0 - deg;
+    double correction = m_correctionFactor - deg;
```
yspolyakov, line 287: Just to confirm. I don't believe we should be using the correctfactor for NATIVE_SIZE=128. Is this what we have in the implementation?

src/pke/include/scheme/ckksrns/ckksrns-fhe.h
sarojaduality: I used a class variable because we input correctionFactor in EvalBootstrapSetup, but we use it in EvalBootstrap.

benchmark/src/ckks-bootstrapping-precision.cpp
sarojaduality: Moved.

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
sarojaduality: Ran all the experiments for FLEXIBLEAUTO, and they give exactly the same best values for the correction factor as FLEXIBLEAUTOEXT. We also see similar values for precision though (20 bits for full packing with ring dimension 2^16), so I'm not sure if this makes sense.

src/pke/extras/ckks-bootstrapping-precision.cpp
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    avrg = std::sqrt(avrg) / result.size();  // get the average
+    return std::abs(std::log2(avrg));
+}
+
+int main(int argc, char* argv[]) {
+#if NATIVEINT == 64
+    size_t numIterations           = 1;
+    std::vector<uint32_t> slotsVec = {32768};
```
yspolyakov, line 65: Throws an exception because # slots > n/2 for the current settings

src/pke/extras/ckks-bootstrapping-precision.cpp
```diff
@@ -0,0 +1,143 @@
...
+    for (uint32_t numSlots : slotsVec) {
+        for (double i = 8; i < 16; i++) {
+            std::cout << "=======================================================================" << std::endl;
+            std::cout << "Number of slots: " << numSlots << std::endl;
+            std::cout << "Correction Factor: " << i << std::endl;
+            double precision = 0.0;
+            for (size_t j = 0; j < numIterations; j++) {
+                precision += MeasureBootstrapPrecision(numSlots, 0);
```
yspolyakov, line 73: I suggest changing the the correction factor from 0 to i. Otherwise, the output is confusing.

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
yspolyakov: I examined this further and ran experiments for `FLEXIBLEAUTOEXT`, `FLEXIBLEAUTO`, and `FIXEDAUTO`. You are correct that there is no difference in the bootstrapping precision between `FLEXIBLEAUTOEXT` and `FLEXIBLEAUTO`. In this case, the deterministic error due to the approximation $sin(x) \approx x$ is much higher than the nondeterministic error due to RLWE noise. All `FLEXIBLEAUTOEXT` does is reduces the fresh encryption error to the rescaling error, but the accumulated nondeterministic error turns out to be lower than then the deterministic sine wave approximation error incurred in bootstrapping.

Note that `FIXEDAUTO` has a significantly lower precision.

dsuponitskiy (approved):
Looks good to me

src/pke/extras/ckks-bootstrapping-precision.cpp
sarojaduality: Updated to valid numbers of slots.

PR #208, opened 2022-10-14
Static Library Linking Support in Cmake

djpiper28:
this allows for cmake add_subdirectory to be used with this project


djpiper28:
A bit more context. This package does not exist on the AUR and I need to use it in my dissertation work. I have a few options: adding it to the AUR and linking the shared library or, as I am doing, adding a subdirectory in cmake then statically linking.

As I am looking at comparing multiple FHE libraries I would prefere not to statically link them but as I am aiming at targetting an embedded device soon I thought that static linking could be better. So this PR is still a WIP but, will allow for static linking within cmake by `add_subdirectory(openfhe)` then `target_link_libraries(yourname openfhe::core)` or similar.

djpiper28:
Cheers for having an initial look at this @yspolyakov, I was wondering if you could give me a bit of advise on the best way to document these changes.

PR #214, opened 2022-10-19
Improving CKKS bootstrapping precision

### LINKED ISSUE #193: Improve the precision of CKKS bootstrapping

sarojaduality:
Closes #193.


src/pke/examples/iterative-ckks-bootstrapping.cpp
```diff
@@ -0,0 +1,193 @@
...
+    parameters.SetScalingModSize(dcrtBits);
+    parameters.SetScalingTechnique(rescaleTech);
+    parameters.SetFirstModSize(firstMod);
+
+    std::vector<uint32_t> levelBudget = {3, 3};
+    uint32_t approxBootstrapDepth     = 9;
+    std::vector<uint32_t> bsgsDim     = {0, 0};
+
+    uint32_t levelsUsedBeforeBootstrap = 10;
+    usint depth =
```
yspolyakov, line 113: This estimate appears to be inaccurate. When I check the output after two iterations, I get 9 available levels instead of expected 10.

src/pke/examples/iterative-ckks-bootstrapping.cpp
```diff
@@ -0,0 +1,193 @@
...
+double CalculateApproximationError(const std::vector<std::complex<double>>& result,
+                                   const std::vector<std::complex<double>>& expectedResult) {
+    if (result.size() != expectedResult.size())
+        OPENFHE_THROW(config_error, "Cannot compare vectors with different numbers of elements");
+
+    // using the Euclidean norm
+    double maxError = 0;
+    for (size_t i = 0; i < result.size(); ++i) {
+        double error = std::abs(result[i].real() - expectedResult[i].real());
+        if (maxError < error)
```
yspolyakov, line 69: Looks like infinity norm to me, not Euclidian norm.

sarojaduality: They use infinity norm in the Meta-BTS paper. Why switch to Euclidean norm?

src/pke/examples/iterative-ckks-bootstrapping.cpp
sarojaduality: Confirmed and changed to 9.

src/pke/examples/iterative-ckks-bootstrapping.cpp
sarojaduality: Changed the comment to say infinity norm.

src/pke/examples/iterative-ckks-bootstrapping.cpp
yspolyakov: I don't think this is the right fix. The meta-bts bootstrapping requires one extra level on top of regular ckks bootstrapping. I did not check it. But the depth of bootstrapping ciphertext is probably now 8 instead of the expected 9.

dsuponitskiy (approved):
Looks good to me

src/pke/examples/iterative-ckks-bootstrapping.cpp
sarojaduality: For some reason, I'm getting 10 available levels now. I updated approxBootstrapDepth to be 8 + (numIterations + 1).

yspolyakov: Works correctly now.

yspolyakov (approved):

PR #254, opened 2022-12-02
Patch to update the dev branch with commits from main

dsuponitskiy:


jbates-duality:
So, mostly what I see here is code of the following categories.  This code is being backported from main to dev, as we've somehow borked the process which keeps patches flowing in one direction from dev->main.  Yuck.

1) What looks like the serialization bugfix being applied to main, needing to be in dev.
1) EMSCRIPTEN catch blocks throughout unittestland that require #ifdef'ing [ew...].
2) EMSCRIPTEN #ifdef blocks involving the fact that it can't do 128-bit integers.  [mmmmkay]
3) the bump of the version number, post v1.0.1 release.  <shrug>
4) various documentation updates.  [bueno]

There is probably more, and I'm going to take another pass through this to make sure that it all looks good.

jbates-duality:
Adding to this:

6. RingGSW changes.  I'm not familiar with what this is doing, so my review is strictly cursory.

jbates-duality:
After staring at this for a bit, I'm not seeing anything that gives me pause.

I have been asked to merge without squash.  I am going to approve and do so now.

jbates-duality (approved):
I am approving this PR.  See my previous comments regarding my thoughts on this patch.

PR #265, opened 2022-12-15
overload MakeCKKSPackedPlaintext with iterators

lolzballs:
Introduces a new overload for CryptoContext::MakeCKKSPackedPlaintext which takes in a pair of iterators. From a users perspective, this allows them to pass in a portion of a vector, instead of having to copy the slice to a new vector.

The extra typename stuff ensures that the iterator's type is double. Not sure if this is really needed since the part that ends up converting them to complex depends on the iterator being of doubles.


sarojaduality:
Thank you for your contribution to OpenFHE! We prefer the code to be as simple as possible, and we believe adding multiple similar APIs may cause confusion. You are welcome to create a function which inputs iterators in your own code, but we plan to exclude it from OpenFHE for the time being. We will consider adding this API in the future if we find that many other users come to us with the same request. Thanks for the time you spent improving OpenFHE, and we always welcome more suggestions, bug fixes, and features that will improve the library.

PR #268, opened 2022-12-18
Set CMAKE_CXX_STANDARD to 17 for user projects

dsuponitskiy:


jbates-duality (approved):
One lineer in CmaleLists.User.txt
Approved.

PR #269, opened 2022-12-18
Removed not used variables

dsuponitskiy:


jbates-duality (approved):
Removal of two unused variables.
Approved.

PR #287, opened 2023-01-27
Moved out MathBackend

### LINKED ISSUE #286: Move GetMathBackendParameters out of backendfxd.h
As in title. It's causing some build issues in the WASM port and since it is only used in one location, why keep it here :)?

iquah1:
Ayoooo as in title

Closes #286


jbates-duality (approved):
This simple change is innocuous, and fixes a compile error when attempting to build a Duality outboard library that uses OpenFHE.  Approving now.

PR #291, opened 2023-01-30
Set CKKSRNS_SCHEME as schemeTag in CKKSPackedEncoding constructors

dsuponitskiy:


src/pke/include/encoding/ckkspackedencoding.h
```diff
@@ -143,7 +143,7 @@ class CKKSPackedEncoding : public PlaintextImpl {
     bool Decode() {
         OPENFHE_THROW(
             not_available_error,
-            "CKKSPackedEncoding::Decode() is not implemented. Use CKKSPackedEncoding::Decode(noiseScaleDeg,scalingFactor,scalTech) instead.");
+            "CKKSPackedEncoding::Decode() is not implemented. Use CKKSPackedEncoding::Decode(...) instead.");
```
sarojaduality, line 146: Why this change?

sarojaduality (commented):
Looks good other than the error message change.

src/pke/include/encoding/ckkspackedencoding.h
dsuponitskiy: the comment is already outdated as bool Decode(size_t depth, double scalingFactor, ScalingTechnique scalTech, ExecutionMode executionMode) has got more arguments (4 instead of 3) now. also it is not a good practice to mention ALL function arguments in a different place as the chance that the comment will get obsolete gets higher over time.

sarojaduality: Got it, makes sense

PR #300, opened 2023-02-04
added check for batch size to be a power of two

### LINKED ISSUE #298: DiscreteFourierTransform::BitReverse gets stuck in endless loop
Hi. 

I have a been playing around with OpenFHE for a bit and noticed that in some cases the method `DiscreteFourierTransform::BitReverse` in [dftransform.cpp](https://github.com/openfheorg/openfhe-development/blob/main/src/core/lib/math/dftransform.cpp) can get stuck in an infinite loop. It seems that `j` and `bit` can both become 0 therefore never breaking the inner loop. 

The following code produces the issue:
```
#include "openfhe.h"

int main(int argc, char const* argv[]) {
  // set params
  lbcrypto::CCParams<lbcrypto::CryptoContextCKKSRNS> params;
  params.SetMultiplicativeDepth(2);
  params.SetScalingModSize(30);
  params.SetBatchSize(10);
  params.SetScalingTechnique(ScalingTechnique::FLEXIBLEAUTO);
  params.SetSecurityLevel(lbcrypto::SecurityLevel::HEStd_128_classic);

  // create context
  auto context = lbcrypto::GenCryptoContext(params);

  // enable features
  context->Enable(PKESchemeFeature::PKE);
  context->Enable(PKESchemeFeature::KEYSWITCH);
  context->Enable(PKESchemeFeature::LEVELEDSHE);

  // geneate keys
  auto keyPair = context->KeyGen();
  context->EvalMultKeyGen(keyPair.secretKey);

  // create ciphertexts
  auto x =
      context->Encrypt(keyPair.publicKey,
                       context->MakeCKKSPackedPlaintext(std::vector<double>{
                           0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0}));
  return 0;
}
```

kimandrik:
fixes #298 
Added check for equality of batch size to a power of two

### DIFF
--- a/src/pke/lib/scheme/bfvrns/bfvrns-parametergeneration.cpp
+++ b/src/pke/lib/scheme/bfvrns/bfvrns-parametergeneration.cpp
@@ -362,6 +362,10 @@ bool ParameterGenerationBFVRNS::ParamsGenBFVRNS(std::shared_ptr<CryptoParameters
     const EncodingParams encodingParams = cryptoParamsBFVRNS->GetEncodingParams();
     if (encodingParams->GetBatchSize() > n)
         OPENFHE_THROW(config_error, "The batch size cannot be larger than the ring dimension.");
+
+    if (encodingParams->GetBatchSize() & (encodingParams->GetBatchSize() - 1))
+        OPENFHE_THROW(config_error, "The batch size can only be set to zero (for full packing) or a power of two.");
+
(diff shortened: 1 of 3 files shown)


dsuponitskiy (approved):
Looks good, started 6 CI jobs. Will merge after they succeed

Issue #305, opened 2023-02-08
Edge-case in multiple-iteration bootstrapping


iquah1:
If we modify [iterative-ckks-bootstrapping.cpp](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/examples/iterative-ckks-bootstrapping.cpp#L155) from 

`//    Plaintext ptxt = cryptoContext->MakeCKKSPackedPlaintext(x, 1, depth - 1, nullptr, numSlots);`

to 

`    Plaintext ptxt = cryptoContext->MakeCKKSPackedPlaintext(x, 1, 0, nullptr, numSlots);`

we get the error of `Bootstrapping number of RNS moduli 11 must be greater than initial number of RNS moduli 31`

sarojaduality:
The scenario Yuriy mentioned in our meeting was if you start with 5 levels remaining, then you should be able to bootstrap up until 10 levels. This scenario in iterative-ckks-bootstrapping.cpp works (set the depth to `depth - 5`).

The one you're describing above is when you start with 31 levels. Bootstrapping would reduce your # levels to 10, so I'm not sure why you would want to bootstrap in this case. @yspolyakov can you confirm that we want to bootstrap here instead of throwing an exception?

iquah1:
I thought the crux of the issue was that the error message isn't particularly helpful? I might be misunderstanding this.

sarojaduality:
Btw calling EvalBootstrap normally without iterations on 31 levels **does perform the bootstrap** on the ciphertext to 10 levels. So either way for this scenario, we should make the with/without iterative bootstrapping cases have the same behavior.

sarojaduality:
Discussed it with Yuriy, and the plan is to change the logic to return the original ciphertext if it is has a higher number of levels than what we would get after bootstrapping.

Issue #229, opened 2022-11-10
Warn if we're doing bootstrapping prematurely


iquah1:
So we don't have a check to determine if we're calling bootstrapping prematurely. Is it possible to add one? E.g I had something with `levels == 7` and I then called bootstrapping but the levels went up to `21`. Is there a way to know pre-bootstrapping if our levels will go up?

PR #314, opened 2023-03-01
Return original ciphertext in bootstrapping if we already start with enough towers.

sarojaduality:
This resolves Issue #305 and Issues #229. If we input a ciphertext with towers > # towers we end up with after bootstrapping, we return the original ciphertext. We exhibit this behavior for both single and double-iteration bootstrapping.

### DIFF
--- a/src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
+++ b/src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
@@ -248,14 +248,14 @@ Ciphertext<DCRTPoly> FHECKKSRNS::EvalBootstrap(ConstCiphertext<DCRTPoly> ciphert
     double timeDecode(0.0);
 #endif
 
-    auto cc     = ciphertext->GetCryptoContext();
-    uint32_t M  = cc->GetCyclotomicOrder();
-    uint32_t L0 = cryptoParams->GetElementParams()->GetParams().size();
+    auto cc        = ciphertext->GetCryptoContext();
+    uint32_t M     = cc->GetCyclotomicOrder();
+    uint32_t L0    = cryptoParams->GetElementParams()->GetParams().size();
+    auto initSizeQ = ciphertext->GetElements()[0].GetNumOfElements();
 
     if (numIterations > 1) {
         // Step 1: Get the input.
         uint32_t powerOfTwoModulus = 1 << precision;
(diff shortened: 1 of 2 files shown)


src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
```diff
@@ -276,9 +276,7 @@ Ciphertext<DCRTPoly> FHECKKSRNS::EvalBootstrap(ConstCiphertext<DCRTPoly> ciphert
         auto ctBootstrappedScaledDown = ctInitialBootstrap->Clone();
         auto bootstrappingSizeQ       = ctBootstrappedScaledDown->GetElements()[0].GetNumOfElements();
         if (bootstrappingSizeQ <= initSizeQ) {
-            OPENFHE_THROW(config_error, "Bootstrapping number of RNS moduli " + std::to_string(bootstrappingSizeQ) +
-                                            " must be greater than initial number of RNS moduli " +
-                                            std::to_string(initSizeQ));
+            return ciphertext->Clone();
```
iquah1, line 279: Out of curiosity why are we doing a clone here when we used to raise an exception? Is it to maintain parity with the behavior below?

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
```diff
@@ -612,6 +610,11 @@ Ciphertext<DCRTPoly> FHECKKSRNS::EvalBootstrap(ConstCiphertext<DCRTPoly> ciphert
     std::cout << "Decoding time: " << timeDecode / 1000.0 << " s" << std::endl;
 #endif
 
+    auto bootstrappingNumTowers = ctxtDec->GetElements()[0].GetNumOfElements();
```
iquah1, line 613: Nit: add a comment describing why we're doing a return here

iquah1 (commented):
LGTM but just some small things:

should we include a comment in the code describing what is happening? I see it present in the [tests](https://github.com/openfheorg/openfhe-development/pull/314/files#diff-09394beba4c96872322a2542048050e2b4c6544d7b331856060b8b385aa18e79R546)

iquah1 (approved):
:rocket:

dsuponitskiy (approved):
Looks good

PR #322, opened 2023-03-02
Fixed vectorization bug

sararv22:
Fixed vectorization in pubkeygen for xor performance difference in ske vs pke


src/binfhe/lib/lwe-pke.cpp
```diff
@@ -69,15 +69,15 @@ LWEKeyPair LWEEncryptionScheme::KeyGenPair(const std::shared_ptr<LWECryptoParams
 // size is the ring dimension N, modulus is the large Q used in RGSW encryption of bootstrapping.
 LWEPublicKey LWEEncryptionScheme::PubKeyGen(const std::shared_ptr<LWECryptoParams> params,
                                             ConstLWEPrivateKey skN) const {
-    int size              = params->GetN();
+    uint32_t size              = params->GetN();
```
dsuponitskiy, line 72: 1. it is a very bad practice to have a variable names "size". It should be renamed. Although, I see that it is not your code originally.
2. this variable's type should be size_t. there should not be any problem, even though GetN() returns uint32_t.

dsuponitskiy (approved):
looks good to me.

PR #327, opened 2023-03-10
MATHBACKEND cleanup

dsuponitskiy:


dsuponitskiy:
All 6 CI jobs were successful.

jbates-duality (commented):
This simply removes the default setting for MATHBACKEND, which requires the build system to specify it.  This was an issue that caused a lot of pain when the defaults for OpenFHE and a system built on top of it did not match in their defaults.

jbates-duality (approved):
Approving now.

PR #336, opened 2023-03-20
278 lmkcdey

Yongwoo-Lee-ccl:
The pull request implements the bootstrapping of an FHEW-like scheme, based on LMKCDEY (ia.cr/2022/198). 

LMKCDEY requires two different types of keys: n RGSW(X^si) and w+1 (~10) Automorphism keys.
Due to the complexity of making a distinguished member of class, I have used std::make_shared<RingGSWACCKeyImpl>(1, 2, n) for both keys ([0, 0, 0:n] for RGSW keys and [0, 0, 0:w+1] for Automorphism keys) which has resulted in a small inefficiency.

Additionally, round-to-odd has not been implemented, as it would require significant changes to the code structure. Instead, the implementation uses the nearest odd integer from the given input, which may slightly increase the noise.

Lastly, I have hardcoded the window parameter w as 10, which is efficient enough while maintaining a small key size.

I have thoroughly tested my changes and ensured that they do not introduce any new issues or conflicts. 

I would greatly appreciate it if you could review my pull request at your earliest convenience. 
Please let me know if there are any further changes or modifications that I can make to improve the code or address any issues.

Thank you for your time and consideration. I look forward to your response.


Yongwoo-Lee-ccl:
Duplicate of #338

PR #357, opened 2023-04-11
Fix eval-decomp example returning wrong output

sararv22:
@RaindropFamily: The eval-decomp example in binfhe was giving the wrong output (for the last digit). It doesn't look like the final bootstrapping in the eval-decomp function is needed (the algorithm in the paper doesn't have it either). Removing the final bootstrapping in evaldecomp function and changing p=8 from p =2 in decrypt for the last digit in the example fixes the issue. Can you review the changes?

### DIFF
--- a/src/binfhe/examples/eval-decomp.cpp
+++ b/src/binfhe/examples/eval-decomp.cpp
@@ -84,7 +84,7 @@ int main() {
         ct1 = decomp[i];
         LWEPlaintext result;
         if (i == decomp.size() - 1) {
-            p = 2;
+            p = 8;
         }
         cc.Decrypt(sk, ct1, &result, p);
         std::cout << "(" << result << " * " << cc.GetMaxPlaintextSpace() << "^" << i << ")";
--- a/src/binfhe/examples/eval-decomp.cpp
+++ b/src/binfhe/examples/eval-decomp.cpp
@@ -56,7 +56,7 @@ int main() {
 
     int q      = 4096;                                               // q
     int factor = 1 << int(logQ - log2(q));                           // Q/q
-    int P      = cc.GetMaxPlaintextSpace().ConvertToInt() * factor;  // Obtain the maximum plaintext space
+    uint64_t P = cc.GetMaxPlaintextSpace().ConvertToInt() * factor;  // Obtain the maximum plaintext space
 
     // Sample Program: Step 2: Key Generation
     // Generate the secret key
@@ -78,15 +78,15 @@ int main() {
     auto decomp = cc.EvalDecomp(ct1);
 
     // Sample Program: Step 5: Decryption
-    auto p = cc.GetMaxPlaintextSpace().ConvertToInt();
+    uint64_t p = cc.GetMaxPlaintextSpace().ConvertToInt();
     std::cout << "Decomposed value: ";
     for (size_t i = 0; i < decomp.size(); i++) {
         ct1 = decomp[i];
         LWEPlaintext result;
         if (i == decomp.size() - 1) {
             // after every evalfloor, the least significant digit is dropped so the last modulus is computed as log p = (log P) mod (log GetMaxPlaintextSpace)
-            int logp = static_cast<int>(log2(P)) % static_cast<int>(log2(p));
-            p        = 1 << logp;
+            auto logp = GetMSB(P - 1) % GetMSB(p - 1);
+            p         = 1 << logp;
         }
         cc.Decrypt(sk, ct1, &result, p);
         std::cout << "(" << result << " * " << cc.GetMaxPlaintextSpace() << "^" << i << ")";
--- a/src/binfhe/examples/eval-decomp.cpp
+++ b/src/binfhe/examples/eval-decomp.cpp
@@ -56,7 +56,7 @@ int main() {
 
     int q      = 4096;                                               // q
     int factor = 1 << int(logQ - log2(q));                           // Q/q
-    int p      = cc.GetMaxPlaintextSpace().ConvertToInt() * factor;  // Obtain the maximum plaintext space
+    int P      = cc.GetMaxPlaintextSpace().ConvertToInt() * factor;  // Obtain the maximum plaintext space
 
     // Sample Program: Step 2: Key Generation
     // Generate the secret key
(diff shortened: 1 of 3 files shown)


yspolyakov:
Make the following changes:
1. Instead hard-coding p = 8, in the example compute log p = log P mod log (GetMaxPlaintexSpace())
2. Update both unit tests for `EvalDecom` to check all digits (rather than the second LSD).

RaindropFamily:
Hello, sorry that I missed the email... Please feel free to pin me if needed.

Sry I don't quite understand why the result is incorrect. Here's the result I get from the example 
`Decomposed value: (1 * 16^0) + (0 * 16^1) + (0 * 16^2) + (1 * 16^3)`
which is as expected.

I have made some more tests using the following loop:
```
for (int i = 0; i < 8; i++) {
        // We first encrypt with large Q
        auto ct1 = cc.Encrypt(sk, p / 2 + i*8 - 1, FRESH, p, Q);

        // Get the MSB
        auto decomp = cc.EvalDecomp(ct1, Q);
        p = cc.GetMaxPlaintextSpace().ConvertToInt();
        std::cout << "Decomposed value: ";
        for (size_t i = 0; i < decomp.size(); i++) {
            ct1 = decomp[i];
            LWEPlaintext result;
            if (i == decomp.size() - 1) {
                p = 2;
            }
            cc.Decrypt(sk, ct1, &result, p, q);
            std::cout << "(" << result << " * " << cc.GetMaxPlaintextSpace() << "^" << i << ")";
            if (i != decomp.size() - 1) {
                std::cout << " + ";
            }
        }
        std::cout << std::endl;
        p = cc.GetMaxPlaintextSpace().ConvertToInt() * factor;
    }
```
and got this result, which is also as expected:
```
Decomposed value: (15 * 16^0) + (15 * 16^1) + (15 * 16^2) + (0 * 16^3)
Decomposed value: (7 * 16^0) + (0 * 16^1) + (0 * 16^2) + (1 * 16^3)
Decomposed value: (15 * 16^0) + (0 * 16^1) + (0 * 16^2) + (1 * 16^3)
Decomposed value: (7 * 16^0) + (1 * 16^1) + (0 * 16^2) + (1 * 16^3)
Decomposed value: (15 * 16^0) + (1 * 16^1) + (0 * 16^2) + (1 * 16^3)
Decomposed value: (7 * 16^0) + (2 * 16^1) + (0 * 16^2) + (1 * 16^3)
Decomposed value: (15 * 16^0) + (2 * 16^1) + (0 * 16^2) + (1 * 16^3)
Decomposed value: (7 * 16^0) + (3 * 16^1) + (0 * 16^2) + (1 * 16^3)
```

Do you mean that you would not like to have the MSB encoded under p = 2?
I thought that was more of an application choice so I used that way to encode.

yspolyakov:
> Sry I don't quite understand why the result is incorrect. Here's the result I get from the example
> `Decomposed value: (1 * 16^0) + (0 * 16^1) + (0 * 16^2) + (1 * 16^3)`
> which is as expected.

This MSD in the output is incorrect. The number being decomposed is 16385, which is 1 + 4*16^3. So it should be 4 instead of 1.

 log p for the MSD should be set to  log P mod log (GetMaxPlaintexSpace()).

This bug has already been fixed by @sararv22

src/binfhe/examples/eval-decomp.cpp
```diff
@@ -70,21 +70,22 @@ int main() {
...
-    p = cc.GetMaxPlaintextSpace().ConvertToInt();
+    auto p = cc.GetMaxPlaintextSpace().ConvertToInt();
     std::cout << "Decomposed value: ";
     for (size_t i = 0; i < decomp.size(); i++) {
         ct1 = decomp[i];
         LWEPlaintext result;
         if (i == decomp.size() - 1) {
-            p = 2;
+            // after every evalfloor, the least significant digit is dropped so the last modulus is computed as log p = (log P) mod (log GetMaxPlaintextSpace)
+            p = pow(2, static_cast<int>(log2(P)) % static_cast<int>(log2(p)));
```
yspolyakov, line 88: It is not a good practice to use floating-point arithmetic for bit operations. It is better to use bitwise operations on integers instead.

RaindropFamily:
> > Sry I don't quite understand why the result is incorrect. Here's the result I get from the example
> > `Decomposed value: (1 * 16^0) + (0 * 16^1) + (0 * 16^2) + (1 * 16^3)`
> > which is as expected.
> 
> The MSD in the output is incorrect. The number being decomposed is 16385, which is 1 + 4*16^3. So it should be 4 instead of 1.
> 
> log p for the MSD should be set to log P mod log (GetMaxPlaintexSpace()).
> 
> This bug has already been fixed by @sararv22

I think what you mean. I think we have a different understanding.
My implementation intentionally made the MSD be 16385 which was why p = 2 instead of p = 8.
That was also why I asked if it was the case that you didn't like p = 2 for MSD.
But anyway I think that's more of an application-layer preference. Both seem fine to me.
It's good that it's solved. Thanks!

RaindropFamily:
> > > Sry I don't quite understand why the result is incorrect. Here's the result I get from the example
> > > `Decomposed value: (1 * 16^0) + (0 * 16^1) + (0 * 16^2) + (1 * 16^3)`
> > > which is as expected.
> > 
> > 
> > The MSD in the output is incorrect. The number being decomposed is 16385, which is 1 + 4*16^3. So it should be 4 instead of 1.
> > log p for the MSD should be set to log P mod log (GetMaxPlaintexSpace()).
> > This bug has already been fixed by @sararv22
> 
> I think what you mean. I think we have a different understanding. My implementation intentionally made the MSD be 16385 which was why p = 2 instead of p = 8. That was also why I asked if it was the case that you didn't like p = 2 for MSD. But anyway I think that's more of an application-layer preference. Both seem fine to me. It's good that it's solved. Thanks!

Sorry my bad…
Yes you are correct. Thanks for clarifying and fixing that!

src/binfhe/examples/eval-decomp.cpp
yspolyakov: @sararv22 Please fix this so I could merge the PR

yspolyakov (approved):

Issue #280, opened 2023-01-19
Multiplicative depth limitation for HPS, HPSPOVERQ and HPSOVERQLEVELED modes of BFV


yspolyakov:
If I change the multiplicative technique from `HPS` to `HPSPOVER` or `HPSPOVERQLEVELED`, the example at https://github.com/openfheorg/openfhe-integer-examples/blob/master/src/strsearch_enc_1.cpp stops working correctly.  The `BEHZ` mode works fine.

sarojaduality:
I replicated the bug in modes HPS, HPSPOVERQ, and HPSPOVERQLEVELED with a simpler example in the branch: Erabelli_280-bfv-mult-bug in the example [bfv-mult-bug.cpp](https://github.com/openfheorg/openfhe-development/blob/Erabelli_280-bfv-mult-bug/src/pke/examples/bfv-mult-bug.cpp).

The noise estimate after decryption gives a value of inf. Here are some of the other noise estimates and parameters:
sigma: 3.19
Berr: 19.14
alpha: 36
Bkey: 1
delta: 512
t: 786433
V: 19618.5
dcrtBits: 60
n: 65536

The mode HPS fails with multiplicative depth 67 and higher.
mult depth: 67
sizeQ: 44

The modes HPSPOVERQ and HPSPOVERQLEVELED fail with multiplicative depth 31 and higher.
mult depth: 67
sizeQ: 21

The depth 31 worked, when I increased the value of P slightly, to be greater than Q, but higher depths still failed. Maybe the bug is due to slightly incorrect or permuted moduli in the precomputations. This would add to the noise, but not enough to fail for lower multiplicative depths for some reason.

There are no issues with the BEHZ mode. It worked with depths up until 116, and higher depths get killed on my 32 GB machine.

yspolyakov:
We will most likely need to update the implementation to use the digits decomposition approach from Section 3.3 of https://eprint.iacr.org/2021/204.pdf. The problem is probably related to the 52-bit precision limitation of doubles.

Currently, the HPS mode should not be used for more than 66 levels. HPSPOVERQ and HPSOVERQLEVELED should not be used for depths larger than 30. The workaround is to use BEHZ for these deep computation scenarios.

PR #392, opened 2023-05-07
280 andrey depth limitations hps

kimandrik:
Changed floating point arithmetics to fixed point arithmetics in both: ScaleAndRound function used in multiplication and ScaleAndRound function used in decryption. 

Questions: 
1) Do we need to add tests for HPS and other variants? If so the test could be slow, as they require lot's of levels for BFV.
2) I keep temporary bfv-mult-bug.cpp example from https://github.com/openfheorg/openfhe-development/tree/Erabelli_280-bfv-mult-bug. Do we need some variant of it?
3) In case of decryption should we use previous method with floating point arithmetics, or move to new with fixed point arithmetics?
4) Should we rename functions in dcrtPoly, as they could be confusing? Make renamings as a part of another issue?

### DIFF
--- a/src/core/include/lattice/hal/dcrtpoly-interface.h
+++ b/src/core/include/lattice/hal/dcrtpoly-interface.h
@@ -1154,7 +1154,7 @@ public:
    */
     virtual DerivedType ScaleAndRound(const std::shared_ptr<Params> paramsOutput,
                                       const std::vector<std::vector<NativeInteger>>& tOSHatInvModsDivsModo,
-                                      const std::vector<double>& tOSHatInvModsDivsFrac,
+                                      const std::vector<NativeInteger>& tOSHatInvModsDivsFrac,
                                       const std::vector<DoubleNativeInt>& modoBarretMu) const = 0;
 
     /**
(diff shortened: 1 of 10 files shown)


yspolyakov:
> 1. Do we need to add tests for HPS and other variants? If so the test could be slow, as they require lot's of levels for BFV.

I would add some tests using a small ring dimension, e.g., 16.

> I keep temporary bfv-mult-bug.cpp example from https://github.com/openfheorg/openfhe-development/tree/Erabelli_280-bfv-mult-bug. Do we need some variant of it?

I think we want to create a unit test based on it

> In case of decryption should we use previous method with floating point arithmetics, or move to new with fixed point arithmetics?

Let us set up a conference call and discuss this in more detail.

> Should we rename functions in dcrtPoly, as they could be confusing? Make renamings as a part of another issue?

Better do it as part of another issue

yspolyakov:
We had a meeting on this. Some code/approaches from this PR will be used to address https://github.com/openfheorg/openfhe-development/issues/280 and https://github.com/openfheorg/openfhe-development/issues/421

PR #395, opened 2023-05-09
Added check to make sure that std::map<>::find() returns a valid iterator

dsuponitskiy:


dsuponitskiy:
All CI jobs succeeded.

pascoec (approved):

PR #403, opened 2023-05-11
Fix for segmentation fault in MakeCKKSPackedPlaintext()

dsuponitskiy:
Limited the upper value of the level to encode the plaintext at.

### DIFF
--- a/src/pke/include/cryptocontext.h
+++ b/src/pke/include/cryptocontext.h
@@ -101,15 +101,21 @@ class CryptoContextImpl : public Serializable {
         const auto cryptoParams = std::dynamic_pointer_cast<CryptoParametersRNS>(GetCryptoParameters());
         if (level > 0) {
             // validation of level: We need to compare it to multiplicativeDepth, but multiplicativeDepth is not
-            // readily available. so we get is numModuli and use it for calculations
+            // readily available. so, what we get is numModuli and use it for calculations
             size_t numModuli = cryptoParams->GetElementParams()->GetParams().size();
             uint32_t multiplicativeDepth =
                 (cryptoParams->GetScalingTechnique() == FLEXIBLEAUTOEXT) ? (numModuli - 2) : (numModuli - 1);
-            if (level > multiplicativeDepth) {
-                std::string errorMsg(
-                    "The level value should be less than or equal to multiplicativeDepth. Currently: level is [" +
-                    std::to_string(level) + "] and multiplicativeDepth is [" + std::to_string(multiplicativeDepth) +
-                    "]");
+            // we throw an exception if level >= numModuli. however, we use multiplicativeDepth in the error message,
+            // so the user can understand the error more easily.
+            if (level >= numModuli) {
+                std::string errorMsg;
+                if (cryptoParams->GetScalingTechnique() == FLEXIBLEAUTOEXT)
+                    errorMsg = "The level value should be less than or equal to (multiplicativeDepth + 1).";
+                else
+                    errorMsg = "The level value should be less than or equal to multiplicativeDepth.";
+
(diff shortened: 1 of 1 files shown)


src/pke/include/cryptocontext.h
```diff
@@ -169,6 +179,17 @@ class CryptoContextImpl : public Serializable {
     Plaintext MakePlaintext(const PlaintextEncodings encoding, const std::vector<int64_t>& value, size_t depth,
                             uint32_t level) const {
         const auto cryptoParams = std::dynamic_pointer_cast<CryptoParametersRNS>(GetCryptoParameters());
+        // validation of level: We need to compare it to multiplicativeDepth, but multiplicativeDepth is not
+        // readily available. what we can get is numModuli, and numModuli is (multiplicativeDepth+1)
+        size_t numModuli             = cryptoParams->GetElementParams()->GetParams().size();
+        uint32_t multiplicativeDepth = numModuli - 1;
```
sarojaduality, line 185: This equation is not true for BFV. It is only true for BGV and CKKS. However, in all three cases we do need level >= numModuli.

sarojaduality (changes requested):

src/pke/include/cryptocontext.h
yspolyakov: A correction. Level = 0 should be used for BFV. For BGV and CKKS, this logic can be used. However, if the scaling mode is set to FLEXIBLEAUTOEXT, the multiplicative depth = numModuli - 2.

yspolyakov (approved):

src/pke/include/cryptocontext.h
dsuponitskiy: The comments have been addressed

src/pke/include/cryptocontext.h
```diff
@@ -98,10 +98,23 @@ class CryptoContextImpl : public Serializable {
     virtual Plaintext MakeCKKSPackedPlaintextInternal(const std::vector<std::complex<double>>& value,
                                                       size_t noiseScaleDeg, uint32_t level,
                                                       const std::shared_ptr<ParmType> params, usint slots) const {
-        Plaintext p;
         const auto cryptoParams = std::dynamic_pointer_cast<CryptoParametersRNS>(GetCryptoParameters());
-        double scFact;
+        if (level > 0) {
+            // validation of level: We need to compare it to multiplicativeDepth, but multiplicativeDepth is not
+            // readily available. so we get is numModuli and use it for calculations
```
sarojaduality, line 104: Actually, what we want to compare it to is numModuli, and it should be if `(level >= numModuli)`. We use multiplicativeDepth so the user can understand the error message more easily. In the case of `FLEXIBLEAUTOEXT`, the check still needs to be `if (level >= numModuli)`, which translates to `if (level > multiplicativeDepth + 1)`

sarojaduality: @yspolyakov Wanted to double check the following:
In both cases, we want to throw an exception if `level >= numModuli`. In FLEXIBLEAUTOEXT, we have `multiplicativeDepth = numModuli - 2` and otherwise we have `multiplicativeDepth = numModuli - 1`. So the error message in the case of FLEXIBLEAUTOEXT should be `The level value should be less than or equal to multiplicativeDepth + 1`, and otherwise the error message is  `The level value should be less than or equal to multiplicativeDepth`.

yspolyakov: I agree with @sarojaduality. We should thrown an exception when `level >= numModuli`. My initial recommendation was incorrect.

dsuponitskiy:
All CI jobs succeeded

PR #415, opened 2023-05-22
Fix some undefined behavior

ericastor:
Ptch out one instance of an index-out-of-bounds error in the BGVRNS scheme, and two instances of casting a potentially-infinite floating point value to int32_t.


yspolyakov:
@ericastor Could you create a branch in `openfheorg-development` from `dev` and resubmit the PR? I just gave you the appropriate access to the repository.

Issue #393, opened 2023-05-07
Adding a large scalar does not work in CKKS


yspolyakov:
When `EvalAdd` + $c$ is called for a large scalar $c$ where than $2^p*c$ is larger than $2^{64}$, the result becomes incorrect due to an overflow for 64-bit numbers. The issue is described in more detail in https://openfhe.discourse.group/t/evaladd-does-not-add-in-some-cases/517/8

PR #436, opened 2023-06-12
Fixes bugs with large scalars in CKKS

### LINKED ISSUE #385: When the scaling method is set to FLEXIBLEAUTOEXT (Default), repeated additions produce incorrect result.
Based off of [[EvalAdd does not add in some cases](https://openfhe.discourse.group/t/evaladd-does-not-add-in-some-cases/517)](https://openfhe.discourse.group/t/evaladd-does-not-add-in-some-cases/517)

```
#include "openfhe.h"
#include <vector>
#include <iostream>

using namespace lbcrypto;
using namespace std;

int main(){

  uint32_t multDepth = 50;
  uint32_t scaleModSize = 59;
  CCParams<CryptoContextCKKSRNS> parameters;
  parameters.SetMultiplicativeDepth(multDepth);
  parameters.SetScalingModSize(scaleModSize);
  CryptoContext<DCRTPoly> cc = GenCryptoContext(parameters);
  cc->Enable(PKE);
  cc->Enable(LEVELEDSHE);
  auto keys = cc->KeyGen();
  auto val = cc->Encrypt(keys.publicKey, cc->MakeCKKSPackedPlaintext(vector < double > {1}));
  for (int i = 0; i < 41; i++) {
    auto added = cc->EvalAdd(val, i);
    Plaintext p;
    cc->Decrypt(keys.secretKey, added, &p);
    p->SetLength(1);
    cout << "Added " << to_string(i) << ":\t" << p << endl;
  }
}
```

You'll see that after the first few iterations (which produce the correct result), we start encountering the wrong result

yspolyakov:
Fixes #385 and #393 
* Fixes the logic for adding large doubles in CKKS
* Fixes the logic for multiplying by large doubles in CKKS
* Adds unit tests for addling large doubles

### DIFF
--- a/src/pke/include/constants.h
+++ b/src/pke/include/constants.h
@@ -152,7 +152,7 @@ enum PlaintextEncodings {
 std::ostream& operator<<(std::ostream& s, PlaintextEncodings p);
 
 enum LargeScalingFactorConstants {
-    MAX_BITS_IN_WORD = 62,
+    MAX_BITS_IN_WORD = 61,
     MAX_LOG_STEP     = 60,
 };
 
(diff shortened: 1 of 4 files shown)


dsuponitskiy (approved):
Looks good

PR #439, opened 2023-06-14
Updated standard security tables

sararv22:
updated stdlatticeparms for ternary, error secret distribution with 2^16 ring dimension for classic 128, 192, 256 security and parameter sets for 128, 192, 256 quantum security levels upto 2^15 ring dimension


yspolyakov (approved):

dsuponitskiy (approved):
Looks good to me

Issue #402, opened 2023-05-11
DCRTPoly::ScaleAndRound() and CryptoContext::RecoverSharedKey() generating non-reduced input to forward NTT


pascoec:
I had to add an [extra reduction loop to the beginning of forward NTT](https://github.com/openfheorg/openfhe-development/blob/dev/src/core/include/math/hal/intnat/transformnat-impl.h#L306) after detecting some input values were greater than the modulus.

I tracked the issue down to 2 functions: [DCRTPoly::ScaleAndRound()](https://github.com/openfheorg/openfhe-development/blob/dev/src/core/lib/lattice/hal/default/dcrtpoly.cpp#L2376) and [CryptoContext::RecoverSharedKey(](https://github.com/openfheorg/openfhe-development/blob/dev/src/pke/lib/cryptocontext.cpp#L1030))

[Proposed fix in branch issue-402](https://github.com/openfheorg/openfhe-development/commit/3781ee944ebb154304abfe72ff1e2565e578f3a2)

PR #442, opened 2023-06-15
fix for issue #402

pascoec:


src/core/include/utils/utilities.h
```diff
@@ -135,29 +135,30 @@ inline uint32_t AdditionWithCarryOut(uint64_t a, uint64_t b, uint64_t& c) {
...
 }
 
 // TODO (dsuponit): the name of this function Max64BitValue() is misleading as it returns the largest value
 // that can be converted from double to int64_t and not the max value of int64_t. The function must be renamed!!!
 constexpr int64_t Max64BitValue() {
-    // (2^63-1)-(2^10-1) => 2^63-2^10 - max value that could be rounded to int64_t
-    return static_cast<int64_t>((uint64_t(1) << 63) - (uint64_t(1) << 10));
+    // (2^63 - 2^9 - 1) = 9223372036854775295 is max int64_t value convertable to double and back to int64_t
+    // without overflow, but there is loss of precision, which causes a warning when implicitly casted to double.
+    // (2^63-1)-(2^10-1) = 9223372036854774784 = int64_t(double(9223372036854775295))
```
yspolyakov, line 148: Remove the unnecessary comments from the code. Use the comments feature of github instead.

src/core/include/utils/utilities.h
```diff
@@ -135,29 +135,30 @@ inline uint32_t AdditionWithCarryOut(uint64_t a, uint64_t b, uint64_t& c) {
...
-    // there is an overflow or not:
-    // return (std::abs(d) > Max64BitValue());
-    // TO BE REVIEWED...
-    const double EPSILON = 0.000001;
-
-    return EPSILON < (std::abs(d) - Max64BitValue());
+    // TODO (dsuponit): the name of this function is64BitOverflow() is misleading as it checks if double can be
+    // converted to int64_t. The name should reflect that. something like isConvertableToInt64() The function must be renamed!!!
+
+    // use of EPSILON check here is incorrect as this function asks a simple binary question: is there overflow or not.
```
yspolyakov, line 159: Same issue about comments here.

src/core/include/lattice/hal/default/dcrtpoly-impl.h
```diff
@@ -1926,128 +1926,105 @@ DCRTPolyImpl<VecType> DCRTPolyImpl<VecType>::ScaleAndRound(
...
-            NativeInteger alpha = static_cast<uint64_t>(nu);
-
-            for (usint j = 0; j < sizeO; j++) {
-                DoubleNativeInt curValue = 0;
-
-                const NativeInteger& oj                                  = paramsOutput->GetParams()[j]->GetModulus();
-                const std::vector<NativeInteger>& tOSHatInvModsDivsModoj = tOSHatInvModsDivsModo[j];
-
-                for (usint i = 0; i < sizeI; i++) {
+            NativeInteger alpha = static_cast<BasicInteger>(nu);
```
yspolyakov, line 1942: Change BasicInteger to uint64_t

src/pke/lib/cryptocontext.cpp
```diff
@@ -754,124 +753,65 @@ void CryptoContextImpl<DCRTPoly>::RecoverSharedKey(PrivateKey<DCRTPoly>& sk,
...
-                for (size_t l = 0; l < ring_dimension; l++) {
-                    lagrange_sum_of_elems_poly.at(l) +=
-                        Lagrange_coeffs[client_indexes[i]].GetElementAtIndex(k).at(l).ModMul(
-                            sk_shares[client_indexes[i]].GetElementAtIndex(k).at(l), modq_k);
+            NativePoly lagrange_sum_of_elems_poly(elementParams->GetParams()[k], Format::COEFFICIENT, true);
+            for (usint i = 0; i < client_indexes_size; ++i) {
+                // TODO: should we enable modmul of polys in coefficient format?
+                const auto& coeff = Lagrange_coeffs[i].GetAllElements()[k];
+                const auto& share = sk_shares[client_indexes[i]].GetAllElements()[k];
+                for (size_t d = 0; d < ring_dimension; ++d) {
```
yspolyakov, line 809: Replace the code using NativeVector instead of NativeInteger

src/core/include/lattice/hal/default/dcrtpoly-impl.h
pascoec: Will do. Just note that
`NativeInteger alpha = static_cast<uint64_t>(nu);`
becomes
`NativeInteger alpha = static_cast<BasicInteger>(static_cast<uint64_t>(nu));`
if BasicInteger != uint64_t

pascoec: which could happen at L1995
https://github.com/openfheorg/openfhe-development/blob/de24e7e423284ca03c9d8c7305bec5f591b60bc5/src/core/include/lattice/hal/default/dcrtpoly-impl.h#LL1995C16-L1995C16

src/core/include/lattice/hal/default/dcrtpoly-impl.h
```diff
@@ -1926,128 +1926,105 @@ DCRTPolyImpl<VecType> DCRTPolyImpl<VecType>::ScaleAndRound(
...
-
-        for (usint i = 0; i < sizeI; i++) {
+        for (size_t i = 0; i < sizeI; ++i) {
+            // possible loss of precision if modulus greater than 2^53 + 1
             const NativeInteger& xi = m_vectors[i + inputIndex][ri];
-            nu += tOSHatInvModsDivsFrac[i] * xi.ConvertToInt();
+            nu += tOSHatInvModsDivsFrac[i] * xi.ConvertToDouble();
         }
-
         if (!is64BitOverflow(nu)) {
```
pascoec, line 1994: This is encountered when NATIVEINT == 32, NATIVEINT == 128, or !defined(HAVE_INT128) && NATIVEINT == 64.

There are various subtle issues with current implementation if NATIVEINT == 32 or NATIVEINT == 128. Should we create templated function that performs correct overflow check dependent upon BasicIntegr type?

src/core/include/lattice/hal/default/dcrtpoly-impl.h
yspolyakov: it would be an issue only for NATIVE_SIZE = 32. Currently NATIVE_SIZE = 32 should only be used for binfhe. So probably the change I suggested is not needed.

src/core/include/lattice/hal/default/dcrtpoly-impl.h
yspolyakov: The current assumption is that that BFV should only be used for NATIVE_SIZE=64 (both when 128-bit int is available and not)

src/pke/lib/cryptocontext.cpp
```diff
@@ -754,124 +753,61 @@ void CryptoContextImpl<DCRTPoly>::RecoverSharedKey(PrivateKey<DCRTPoly>& sk,
...
     // vector of indexes of the clients
-    std::vector<int64_t> client_indexes;
-    for (size_t i = 1; i <= N; i++) {
-        if (sk_shares.find(i) != sk_shares.end()) {
+    std::vector<uint32_t> client_indexes;
+    for (uint32_t i = 1; i <= N; ++i) {
+        if (sk_shares.find(i) != sk_shares.end())
             client_indexes.push_back(i);
-        }
+        // add early exit if client_indexes.size() == threshold?
```
yspolyakov, line 760: Resolve this and remove the comment

yspolyakov (approved):

PR #451, opened 2023-06-23
Functions to check enabled features in the cryptocontext

dsuponitskiy:


dsuponitskiy:
All 6 CI jobs succeeded

jbates-duality (approved):
This is straightforward.  Facilitator method to check whether features are enabled, instead of if statements, and additional level of indent.  Code is more readable.  Do not see any other mods.  Approving.

PR #453, opened 2023-06-26
Small correction to the migration guide (from Palisade to OpenFHE)

dsuponitskiy:


jbates-duality (approved):
This is a trivial one-liner in the documentation source.  Approving.

PR #460, opened 2023-06-30
Add instructions for custom build location in user project

dsuponitskiy:
…tion and updated docs with other OSs where OpenFHE has been sucessfully linked


jbates-duality (approved):
Helpful output from CMAKE has been added.  Helpful instruction for alternate install locaitons.  Update of supported OSes - this may be erroneous, and we will review this next week.  I am approving this change.

PR #463, opened 2023-07-04
Problem with two coexisting CKKS crypto contexts with different ring dimensions

dsuponitskiy:


dsuponitskiy:
All CI jobs succeeded

src/core/lib/math/dftransform.cpp
```diff
@@ -45,14 +45,29 @@
...
-size_t DiscreteFourierTransform::m_M                             = 0;
-size_t DiscreteFourierTransform::m_Nh                            = 0;
-bool DiscreteFourierTransform::m_isInitialized                   = false;
+std::unordered_map<uint32_t, DiscreteFourierTransform::PrecomputedValues> DiscreteFourierTransform::precomputedValues;
 
-/// precomputed rotation group indices
-std::vector<uint32_t> DiscreteFourierTransform::m_rotGroup;
-/// precomputed ksi powers
-std::vector<std::complex<double>> DiscreteFourierTransform::m_ksiPows;
+DiscreteFourierTransform::PrecomputedValues::PrecomputedValues(uint32_t m, uint32_t nh) {
```
yspolyakov, line 50: I would change the name of the method from PrecomputedValues to PrecomputeValues

src/core/lib/math/dftransform.cpp
```diff
@@ -211,18 +206,19 @@ std::vector<std::complex<double>> DiscreteFourierTransform::InverseTransform(std
     return invDftRemainder;
 }
 
-void DiscreteFourierTransform::FFTSpecialInvLazy(std::vector<std::complex<double>>& vals) {
+void DiscreteFourierTransform::FFTSpecialInvLazy(std::vector<std::complex<double>>& vals,
```
yspolyakov, line 209: Why is prepValues needed? Why not read it from the static map associated with the class?

src/core/lib/math/dftransform.cpp
dsuponitskiy: It is the constructor for the PrecomputedValues struct.

yspolyakov: Understood

yspolyakov (approved):

yspolyakov:
I see about 2% runtime increase for `CKKSrns_Decryption` in the lib-benchmark as compared to the `dev` branch. Also, there is about 0.5$ runtime increase for `CKKSrns_Encryption`. My guess is that the extra cost is associated to resolving the reference to the STL map item. Please independently run the benchmarks to confirm my results.

PR #466, opened 2023-07-06
Add numAutoKeys to RingGSWCryptoParams

Yongwoo-Lee-ccl:
This PR introduces parameterization for the number of automorphism keys in the LMKCDEY bootstrapping. 
The formerly hardcoded value of `m_window = 10` has been updated and is now integrated into `RingGSWCryptoParams` under the variable name `numAutoKeys`.

### DIFF
--- a/src/binfhe/include/binfhecontext.h
+++ b/src/binfhe/include/binfhecontext.h
@@ -74,8 +74,8 @@ public:
    * @return creates the cryptocontext
    */
     void GenerateBinFHEContext(uint32_t n, uint32_t N, const NativeInteger& q, const NativeInteger& Q, double std,
-                               uint32_t baseKS, uint32_t baseG, uint32_t baseR, uint32_t numAutoKeys = 1,
-                               SECRET_KEY_DIST keyDist = UNIFORM_TERNARY, BINFHE_METHOD method = GINX);
+                               uint32_t baseKS, uint32_t baseG, uint32_t baseR, SECRET_KEY_DIST keyDist = UNIFORM_TERNARY, 
+                               BINFHE_METHOD method = GINX, uint32_t numAutoKeys = 10);
 
     /**
    * Creates a crypto context using custom parameters.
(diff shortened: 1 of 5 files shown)


src/binfhe/include/binfhecontext.h
```diff
@@ -74,7 +74,8 @@ class BinFHEContext : public Serializable {
    * @return creates the cryptocontext
    */
     void GenerateBinFHEContext(uint32_t n, uint32_t N, const NativeInteger& q, const NativeInteger& Q, double std,
```
yspolyakov, line 76: Please move the numAutoKeys parameter to the end. Otherwise, it will break the code using the existing API.

yspolyakov:
Thank you @Yongwoo-Lee-ccl  I have the following suggestions:

1. Please move the numAutoKeys argument to the end in all methods that need it. This is needed for backwards compatibility. Otherwise, the code using the existing API will be broken.
2. As far as I understand, numAutoKeys should be set to 10 by default. I did not see it in the code (I might have missed it). In other words, what will happen if the use tries a parameter set using the default value of 0. My suggestion is to use 10 by default (when the user supplies 0 as the input), unless the user manually enter a different number (not 0).

Yongwoo-Lee-ccl:
@yspolyakov , numAutoKeys is moved to the end in all methods, and initialized to 10.
In addition, it throws when numAutoKeys is zero and method is LMKCDEY (Line 79-81, rgsw-cryptoparameters.h.)

> Thank you @Yongwoo-Lee-ccl I have the following suggestions:
> 
> 1. Please move the numAutoKeys argument to the end in all methods that need it. This is needed for backwards compatibility. Otherwise, the code using the existing API will be broken.
> 2. As far as I understand, numAutoKeys should be set to 10 by default. I did not see it in the code (I might have missed it). In other words, what will happen if the use tries a parameter set using the default value of 0. My suggestion is to use 10 by default (when the user supplies 0 as the input), unless the user manually enter a different number (not 0).

yspolyakov (approved):

yspolyakov:
The code looks good to me. I started a CI job for it, too. @dsuponitskiy please review. If all looks good to you, please merge it.

yspolyakov:
@Yongwoo-Lee-ccl The unit tests for LMKCDEY are failing. Please see https://github.com/openfheorg/openfhe-development/actions/runs/5489284270/jobs/10003907106 for more details.

dsuponitskiy (approved):
looks good

PR #472, opened 2023-07-10
added support for Compress to BFV for HPS* modes

yspolyakov:
As part of the change, rewrote MultiplyAndRound for the native backend. This method is now used for BFV decryption for a DCRTPoly with a single RNS limb.


dsuponitskiy (approved):
Looks good

PR #489, opened 2023-07-17
Hardening and optimizations of CKKS bootstrapping for sparse secret distribution

yspolyakov:
1. Increased the Hamming weight from 64 to 192.
2. Switched from Chebyshev approximation to Chebyshev approximation + double-angle formula.

### DIFF
--- a/src/pke/include/scheme/ckksrns/ckksrns-fhe.h
+++ b/src/pke/include/scheme/ckksrns/ckksrns-fhe.h
@@ -199,8 +199,10 @@ public:
         ar(cereal::base_class<FHERNS>(this));
     }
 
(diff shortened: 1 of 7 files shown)


src/pke/include/scheme/ckksrns/ckksrns-fhe.h
```diff
@@ -199,8 +199,10 @@ class FHECKKSRNS : public FHERNS {
         ar(cereal::base_class<FHERNS>(this));
     }
 
+    // To be deprecated; left for backwards compatibility
```
dsuponitskiy, line 202: just a reminder: we can add an attribute to produce a trace for deprecated functions

dsuponitskiy (approved):

PR #524, opened 2023-08-02
Rename all files related to cryptocontext generation

dsuponitskiy:


jbates-duality (approved):
Straightforward renaming of files and #pragma once-style macro-based include-file-excluder/de-dupers.

PR #527, opened 2023-08-03
patched encoding

andreea-alexandru:
The encoding for complex numbers in scheme switching has now the patches for avoiding overflow from ckksrns-fhe.


dsuponitskiy (approved):
looks good

PR #563, opened 2023-10-02
Added a function to get the call stack

dsuponitskiy:


jbates-duality (approved):
const-correctness in de-mangle routine.  routine to generate a stack trace.  Comparing to code I've used in the past, this looks correct.  There is a minor change in the demange output, but I don't think we're using this in any automated test scripts, so it's OK.  Approving.

Issue #545, opened 2023-08-22
Optimize creation of automorphism keys in bootstrapping and scheme switching


andreea-alexandru:
Currently, the internal (GetScheme()->) EvalAtIndexKeyGen first creates the keys for the given set of indices, and only then removes duplicates that were created before. It should only create new keys for the indices that don't exist.

The functions that need to be changed are EvalBootstrapKeyGen, EvalCKKStoFHEWKeyGen, EvalFHEWtoCKKSKeyGen and EvalSchemeSwitchingSetup.

PR #566, opened 2023-10-13
Create one map for all automorphism keys

dsuponitskiy:


dsuponitskiy:
Some questions to be clarified:
1. [This issue](https://github.com/openfheorg/openfhe-development/issues/545) suggests to change EvalSchemeSwitchingSetup().
2. We may want to serialize/deserialize keys depending on the required functionality

andreea-alexandru:
Typo, it is EvalSchemeSwitchingKeyGen().

yspolyakov:
@dsuponitskiy The optimization is not fully implemented. Please make sure the desired new indices are matched against the indices for existing keys before any eval key generation (the most expensive operation) is called.

PR #567, opened 2023-10-13
Added an additional PATH requirement for user apps and corrected a do…

dsuponitskiy:
…cument note


jbates-duality (approved):
Fixed a minor punctuation issue, per my request.  re-approving.

PR #586, opened 2023-11-13
572 expand serialization for ckks bootstrapping

dsuponitskiy:


dsuponitskiy:
Started 6 CI jobs

src/pke/unittest/utckksrns/UnitTestBootstrap.cpp
```diff
@@ -598,6 +608,88 @@ class UTCKKSRNS_BOOT : public ::testing::TestWithParam<TEST_CASE_UTCKKSRNS_BOOT>
...
+
+            cc->EvalBootstrapPrecompute(testData.slots);
+            cc->EvalBootstrapPrecompute(testData.slots / 2);
+
+            auto keyPair = cc->KeyGen();
+            cc->EvalMultKeyGen(keyPair.secretKey);
+            cc->EvalBootstrapKeyGen(keyPair.secretKey, testData.slots);
+            cc->EvalBootstrapKeyGen(keyPair.secretKey, testData.slots / 2);
+
+            // the number of vector elements should not be greater than the number of slots!!!
```
andreea-alexandru, line 649: remove comment?

andreea-alexandru (approved):
Looks good to me

src/pke/include/scheme/ckksrns/ckksrns-fhe.h
```diff
@@ -111,6 +111,24 @@ class CKKSBootstrapPrecom {
 
     // coefficients corresponding to conj(U0^T); used in encoding
     std::vector<std::vector<ConstPlaintext>> m_U0hatTPreFFT;
+
+    template <class Archive>
+    void save(Archive& ar) const {
+        ar(cereal::make_nvp("dim1_Enc", m_dim1));
```
yspolyakov, line 117: Why don't we store m_paramsEnc[CKKS_BOOT_PARAMS::GIANT_STEP]) in this case?
From looking at the code, it looks like we use m_dim1 only for the linear case (when level budgets are {1,1}). We use m_paramsEnc for other cases (when level budgets > 1). So I think we should serialize both m_dim1 and m_paramsEnc [CKKS_BOOT_PARAMS::GIANT_STEP].

src/pke/include/scheme/ckksrns/ckksrns-fhe.h
```diff
@@ -111,6 +111,24 @@ class CKKSBootstrapPrecom {
...
+        ar(cereal::make_nvp("dim1_Dec", m_paramsDec[CKKS_BOOT_PARAMS::GIANT_STEP]));
+        ar(cereal::make_nvp("slots", m_slots));
+        ar(cereal::make_nvp("lEnc", m_paramsEnc[CKKS_BOOT_PARAMS::LEVEL_BUDGET]));
+        ar(cereal::make_nvp("lDec", m_paramsDec[CKKS_BOOT_PARAMS::LEVEL_BUDGET]));
+    }
+
+    template <class Archive>
+    void load(Archive& ar) {
+        ar(cereal::make_nvp("dim1_Enc", m_dim1));
+        ar(cereal::make_nvp("dim1_Dec", m_paramsDec[CKKS_BOOT_PARAMS::GIANT_STEP]));
```
yspolyakov, line 127: Same issue here

src/pke/unittest/utckksrns/UnitTestBootstrap.cpp
```diff
@@ -288,6 +294,10 @@ static std::vector<TEST_CASE_UTCKKSRNS_BOOT> testCases = {
...
 #endif
     // ==========================================
+    // TestType,           Descr, Scheme,         RDim, MultDepth,  SModSize,     DSize, BatchSz, SecKeyDist,      MaxRelinSkDeg, FModSize,  SecLvl,       KSTech, ScalTech,        LDigits,      PtMod, StdDev, EvalAddCt, KSCt, MultTech, EncTech, PREMode, LvlBudget, Dim1,       Slots
+    { BOOTSTRAP_SERIALIZE, "01", {CKKSRNS_SCHEME, RDIM, MULT_DEPTH, SMODSIZE,     DFLT,  DFLT,    UNIFORM_TERNARY, DFLT,          FMODSIZE,  HEStd_NotSet, HYBRID, FIXEDAUTO,       NUM_LRG_DIGS, DFLT,  DFLT,   DFLT,      DFLT, DFLT,     DFLT,    DFLT},   { 1, 1 },  { 32, 32 }, RDIM/2 },
+    { BOOTSTRAP_SERIALIZE, "02", {CKKSRNS_SCHEME, RDIM, MULT_DEPTH, SMODSIZE,     DFLT,  DFLT,    SPARSE_TERNARY,  DFLT,          FMODSIZE,  HEStd_NotSet, HYBRID, FIXEDAUTO,       NUM_LRG_DIGS, DFLT,  DFLT,   DFLT,      DFLT, DFLT,     DFLT,    DFLT},   { 1, 1 },  { 32, 32 }, RDIM/2 },
+    // ==========================================
```
yspolyakov, line 300: Missing unit tests for level budgets > 1. Add two more tests with {2,2},{0,0}. Then two more tests with {2,2}{4,4}. 4 tests in total. The second {} is for dim1 values.

src/pke/include/scheme/ckksrns/ckksrns-fhe.h
andreea-alexandru: It is redundant to serialize both m_dim1 and m_paramsEnc [CKKS_BOOT_PARAMS::GIANT_STEP], since after deserialization, in EvalBootstrapPrecompute, m_dim1 = dim1[0] is again fed to GetCollapsedFFTParams, which sets m_paramsEnc [CKKS_BOOT_PARAMS::GIANT_STEP] in the same way as it is in EvalBootstrapSetup. The reason why for decoding we serialize m_paramsDec[CKKS_BOOT_PARAMS::GIANT_STEP] is to not create another data member corresponding to dim1[1]. Currently, we serialize the minimum amount of data required for precomputations.

yspolyakov: I agree that they are redundant but the code for the Linear case currently uses dim1 (some local internal refactoring would be needed). Also, I noticed that the unit tests only for levelBudget = {1,1}. So I am not sure the FFT case is even working now. Worth checking first... Is it even working?

yspolyakov: If the unit tests already work, then the extra parameter is not needed.

dsuponitskiy: i added 4 more unit tests. the old and new ones succeeded with and without serializing/deserializing  m_paramsEnc[CKKS_BOOT_PARAMS::GIANT_STEP]

yspolyakov: I just looked at the code. It does look like m_dim1 is used there. So no need to serialize the extra parameter.

dsuponitskiy: let me commit the new unit tests and run all CIs. i am not adding m_paramsEnc to be serialized for now

andreea-alexandru: m_dim1 is serialized from dim1[0] and deserialized, so the correct value is used in the linear case. In the FFT case, m_dim1 = dim1[0] is fed to the same precomputation GetCollapsedFFTParams in both EvalBootstrapSetup and EvalBootstrapPrecompute, so the m_paramsEnc are set correctly.

Alternatively, a redundant but visually simpler way is to only compute m_paramsEnc and m_paramsDec once in EvalBootstrapSetup and serialize them, along with m_dim1, and not recompute in EvalBootstrapPrecompute.

dsuponitskiy: i have the new unit tests committed and started CI jobs

yspolyakov (approved):

PR #587, opened 2023-11-14
Update X(N)OR gate mapping function and delete benchmark and examples of X(N)OR_FAST

HONGDUCK:


Yongwoo-Lee-ccl (commented):
The code is correct.

Yongwoo-Lee-ccl (approved):

Issue #578, opened 2023-11-04
Fix X(N)OR’s mapping functions


HONGDUCK:
```
LWECiphertext BinFHEScheme::EvalBinGate(const std::shared_ptr<BinFHECryptoParams>& params, BINGATE gate,
                                        const RingGSWBTKey& EK, ConstLWECiphertext& ct1,
                                        ConstLWECiphertext& ct2) const {
    if (ct1 == ct2)
        OPENFHE_THROW(config_error, "Input ciphertexts should be independant");

    // By default, we compute XOR/XNOR using a combination of AND, OR, and NOT gates
    if ((gate == XOR) || (gate == XNOR)) {
        const auto& ctAND1 = EvalBinGate(params, AND, EK, ct1, EvalNOT(params, ct2));
        const auto& ctAND2 = EvalBinGate(params, AND, EK, EvalNOT(params, ct1), ct2);
        const auto& ctOR   = EvalBinGate(params, OR, EK, ctAND1, ctAND2);

        // NOT is free so there is not cost to do it an extra time for XNOR
        return (gate == XOR) ? ctOR : EvalNOT(params, ctOR);
    }

    LWECiphertext ctprep = std::make_shared<LWECiphertextImpl>(*ct1);
    // the additive homomorphic operation for XOR/NXOR is different from the other gates we compute
    // 2*(ct1 - ct2) mod 4 for XOR, me map 1,2 -> 1 and 3,0 -> 0
    if ((gate == XOR_FAST) || (gate == XNOR_FAST)) {
        LWEscheme->EvalSubEq(ctprep, ct2);
        LWEscheme->EvalAddEq(ctprep, ctprep);
    }
    else {
        // for all other gates, we simply compute (ct1 + ct2) mod 4
        // for AND: 0,1 -> 0 and 2,3 -> 1
        // for OR: 1,2 -> 1 and 3,0 -> 0
        LWEscheme->EvalAddEq(ctprep, ct2);
    }

    auto acc{BootstrapGateCore(params, gate, EK.BSkey, ctprep)};
    ...
```

I think that this code is implemented like described in <a href="https://eprint.iacr.org/2020/086">https://eprint.iacr.org/2020/086.

When we use X(N)OR gate, we execute (ct1 && ~ct2) || (~ct1 && ct2). This process need 3 blind rotation so it’s slower then AND, OR gate.

If we use X(N)OR_FAST gate, we use following mapping function.
[...]

PR #593, opened 2023-11-16
578 fast xor

Yongwoo-Lee-ccl:
We changed X(N)OR, now it has the same speed and failure probability as other gates.
Thus, X(N)OR_FAST is kept for backwards compatibility, but it points to X(N)OR under the hood. 
We refer #578 for the detail.


yspolyakov:
Thank you for submitting it. I will re-add X(N)OR_FAST (mapping them to regular XOR and XNOR) for backward compatibility. If someone tries to run a circuit that uses FAST flavors, e.g., in Google Transpiler, the code will still work.

PR #597, opened 2023-11-17
Corrected the serialization unittest for CKKS boostrapping

dsuponitskiy:


yspolyakov (approved):

andreea-alexandru (approved):
The order of operations is correct now.

PR #606, opened 2023-11-28
resolve issue 604

pascoec:
- LastPrime(uint32_t nBits, uint32_t m) added to nbtheory.h (Finds the max prime q that satisfies q = 1 mod m with at most nBits bits)
- Replaces instances of PreviousPrime(FirstPrime()) with LastPrime()
- Replaces instances of GetMaxPrime() with LastPrime()
- GetMaxPrime() removed from nbtheory.h
- Where appropriate, replaces instances of NextPrime(FirstPrime()) with PreviousPrime(LastPrime())


src/pke/lib/scheme/bfvrns/bfvrns-cryptoparameters.cpp
```diff
@@ -684,16 +684,15 @@ void CryptoParametersBFVRNS::PrecomputeCRTTables(KeySwitchTechnique ksTech, Scal
...
 
         BigInteger Q(GetElementParams()->GetModulus());
         BigInteger maxConvolutionValue(BigInteger(2 * n) * BigInteger(GetPlaintextModulus()) * Q * Q);
         // check msk is large enough
         while (Q * B * BigInteger(m_msk) < maxConvolutionValue) {
-            auto firstInteger{FirstPrime<NativeInteger>(s + 1, 2 * n)};
-            m_msk = NextPrime<NativeInteger>(firstInteger, 2 * n);
-            if (++s >= 60)
+            m_msk = LastPrime<NativeInteger>(++s, 2 * n);
+            if (s >= 60)  // TODO: MAX_MODULUS_SIZE?
```
yspolyakov, line 695: I think this check should be done before calling `LastPrime`

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
```diff
@@ -198,8 +198,9 @@ std::pair<std::vector<NativeInteger>, uint32_t> ParameterGenerationBGVRNS::compu
             "Change parameters! Try reducing the number of additions per level, "
             "number of key switches per level, or the digit size. We cannot support moduli greater than 60 bits.");
     }
+
     uint32_t totalModSize = firstModSize;
-    moduliQ[0]            = FirstPrime<NativeInteger>(firstModSize, cyclOrder);
+    moduliQ[0]            = FirstPrime<NativeInteger>(firstModSize - 1, cyclOrder);
```
yspolyakov, line 203: I think this should be `LastPrime` with firstModSize

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
```diff
@@ -213,11 +214,13 @@ std::pair<std::vector<NativeInteger>, uint32_t> ParameterGenerationBGVRNS::compu
                 "Change parameters! Try reducing the number of additions per level, "
                 "number of key switches per level, or the digit size. We cannot support moduli greater than 60 bits.");
         }
-        totalModSize += extraModSize;
-        moduliQ[numPrimes] = FirstPrime<NativeInteger>(extraModSize, cyclOrder);
+
+        moduliQ[numPrimes] = FirstPrime<NativeInteger>(extraModSize - 1, cyclOrder);
```
yspolyakov, line 218: Why not LastPrime?

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
```diff
@@ -248,28 +251,28 @@ std::pair<std::vector<NativeInteger>, uint32_t> ParameterGenerationBGVRNS::compu
...
-                moduliQ[1] = NextPrime<NativeInteger>(moduliQ[1], cyclOrder);
+                moduliQ[1] = PreviousPrime<NativeInteger>(moduliQ[1], cyclOrder);
             }
 
             for (size_t i = 2; i < numPrimes; i++) {
-                moduliQ[i] = NextPrime<NativeInteger>(moduliQ[i - 1], cyclOrder);
+                moduliQ[i] = PreviousPrime<NativeInteger>(moduliQ[i - 1], cyclOrder);
                 while (moduliQ[i] == moduliQ[0] || moduliQ[i] == moduliQ[numPrimes] || moduliQ[i] == plainModulusInt) {
-                    moduliQ[i] = NextPrime<NativeInteger>(moduliQ[i], cyclOrder);
+                    moduliQ[i] = PreviousPrime<NativeInteger>(moduliQ[i], cyclOrder);
```
yspolyakov, line 263: I am concerned that this logic with PrevousPrime can go to smaller bit size than requested. This is theoretically possible with BGV. We should probably put a logic to make sure the bitwidth does not become smaller.

yspolyakov: Generally speaking, the moduli should not become smaller. We may need to revisit this logic.

src/pke/lib/scheme/bfvrns/bfvrns-cryptoparameters.cpp
pascoec: Should I remove the check completely?
For NativeInteger, there is a check to see if ++s is greater than MAX_MODULUS_SIZE within the LastPrime call. Behavior for NATIVESIZE=64 would be exactly the same, but s could potentially grow larger than 60 for NATIVESIZE=128.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
pascoec: a few unit tests would request LastPrime(19, 16384), which throws an exception because the returned prime is 18 bits. If LastPrime(19, 16384) is unrealistic, we can change the test parameters.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
pascoec: a few unit tests would request LastPrime(12, 512), LastPrime(10, 512), LastPrime(10, 1024), LastPrime(10, 16384). These all seem unrealistic. We can change the unit tests parameters.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
pascoec: I added checks in the code here to ensure the requested number of bits matches the returned number of bits.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
yspolyakov: We should increase the bit size in unit tests (by 1 or 2, whichever is enough)

src/pke/lib/scheme/bfvrns/bfvrns-cryptoparameters.cpp
yspolyakov: Yes, you can remove it

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
yspolyakov: these are not realistic. Increase the bit sizes in the tests to first ones that pass.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
pascoec: keeping FirstPrime<NativeInteger>(firstModSize - 1, cyclOrder);

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
pascoec: keeping FirstPrime<NativeInteger>(extraModSize - 1, cyclOrder);

yspolyakov (approved):

PR #612, opened 2023-12-05
Validate multDepth as a cryptocontext parameter

dsuponitskiy:


caesaretos (commented):
OpenFHE is unlikely able to support more than ~100-200 multiplicative depths, and we better not claim that OpenFHE supports it (as the message implies). The upper limit 1000 has no scientific basis and better not be printed out to the user.

Something along the following message would be more meaningful.
"The multiplicative depth of _entered_val_ may not be computationally feasible. Please try a smaller depth, such as 30, to improve performance"

Issue #555, opened 2023-09-05
Reduce MIN_SIZE for BFV moduli


yspolyakov:
The current setting of 30 bits does not support the ring dimension of 1024. It should be reduce to a value in the range of 15-20.

yspolyakov:
Set MIN_SIZE to 14 as 12289 is the first prime that is congruent to 1 mod 2*512 (512 is the smallest secure ring dimension)

PR #620, opened 2023-12-15
Fixes a bug in the interface for BFV generation

### LINKED ISSUE #619: BFV uses first mod size instead of scaling mod size when setting a custom size for RNS moduli

yspolyakov:
1. Now uses the scaling mod size instead of first mod size for setting the size of RNS moduli in BFV
2. Reduces the MIN_SIZE from 30 to 14

Closes #619 and #555


dsuponitskiy (approved):
Looks good

PR #626, opened 2023-12-21
Fixed GetBootstrapDepth

yspolyakov:
* Updated `GetBootstrapDepth` to use the correct number of levels (was 1 lower)
* Updated the bootstrapping examples to show the correct number of remaining levels after bootstrapping


dsuponitskiy (approved):
Looks good

PR #628, opened 2023-12-21
Allow only 2-elem ciphertexts for EvalRotation()

dsuponitskiy:


src/pke/include/cryptocontext.h
```diff
@@ -2100,6 +2101,10 @@ class CryptoContextImpl : public Serializable {
    */
     Ciphertext<Element> EvalRotate(ConstCiphertext<Element> ciphertext, int32_t index) const {
         CheckCiphertext(ciphertext);
+        // rotation can be performed on 2-element ciphertexts only
+        if (ciphertext->NumberCiphertextElements() != 2) {
```
yspolyakov, line 2105: Add it at the EvalAutomorphism level instead. This will capture EvalRotate, EvalAtIndex, and other other operations derived from EvalAutomorphism

dsuponitskiy: Addressed. the changes have been committed

yspolyakov (approved):

PR #632, opened 2023-12-29
Update lwe-pke.h

Wangkaixing:
Correct the parameters in the comments of the function KeyGen and KeyGenGaussian


dsuponitskiy:
will be handled by https://github.com/openfheorg/openfhe-development/pull/659

PR #633, opened 2023-12-29
Add `<cmath>` header to fix ambiguity for `abs`

sloede:
While trying to build OpenFHE using [BinaryBuilder.jl](https://github.com/JuliaPackaging/BinaryBuilder.jl), I received some method ambiguity errors for the use of `abs` on macOS, on both `aarch64` ([log](https://buildkite.com/julialang/yggdrasil/builds/7423#018cb78b-6c8b-4a3f-8d6d-fc563f4a950c/630-937)) and `x86_64` ([log](https://buildkite.com/julialang/yggdrasil/builds/7423#018cb78b-6c88-46d2-b894-f44e06259b21/631-934)):
```
In file included from /workspace/srcdir/openfhe-development/src/core/lib/utils/utilities.cpp:36:
/workspace/srcdir/openfhe-development/src/core/include/utils/utilities.h:118:12: error: call to 'abs' is ambiguous
    return std::abs(d) > static_cast<double>(Max64BitValue());
           ^~~~~~~~
/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/include/stdlib.h:129:6: note: candidate function
int      abs(int) __pure2;
         ^
/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/include/c++/v1/stdlib.h:111:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long      abs(     long __x) _NOEXCEPT {return  labs(__x);}
                                           ^
/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/include/c++/v1/stdlib.h:113:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long long abs(long long __x) _NOEXCEPT {return llabs(__x);}
                                           ^
In file included from /workspace/srcdir/openfhe-development/src/core/lib/utils/utilities.cpp:36:
/workspace/srcdir/openfhe-development/src/core/include/utils/utilities.h:135:16: error: call to 'abs' is ambiguous
        return std::abs(d) <= static_cast<double>(std::numeric_limits<int32_t>::max());
               ^~~~~~~~
/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/include/stdlib.h:129:6: note: candidate function
int      abs(int) __pure2;
         ^
[...]

PR #634, opened 2023-12-31
Fix library install directories on Windows

sloede:
Currently, OpenFHE installs all library products in the `lib` directory. AFAIK, this is wrong (or at least bad practice) on Windows, where the `.dll` files belong into the `bin` directory.

This PR remedies this by configuring the library installation in CMake to do The Right Thing:
* On *nix systems, all library products go to `lib`
* On Windows systems (including Cygwin and MinGW), static libraries and `.dll.a` or `.lib` libraries go to `lib`, while the dynamic libraries go to `bin`


sloede:
From https://github.com/openfheorg/openfhe-development/pull/633#issuecomment-1877451241:
> At the moment I am not sure how we are going to proceed with PR #634. I would need to discuss this with my colleague(s) first.

@dsuponitskiy Sure, let me know if I can be of any help. Right now, I don't see any downsides to the proposed change, since it essentially leaves all non-DLL systems unchanged and fixes only the install locations for DLL systems.

Note that the proposed change is also the default for the currently required minimum [CMake version 3.5.2](https://cmake.org/cmake/help/v3.5/command/install.html#targets), while at since [CMake version 3.14](https://cmake.org/cmake/help/v3.14/command/install.html#targets) it is possible to change the install location ad hoc during configure time using the `CMAKE_INSTALL_XXX` variables.

sloede:
@dsuponitskiy Are there any updates on this issue?

arcturusannamalai:
@sloede - you may want to change your code to be conditional to WINDOWS
e.g. see: https://stackoverflow.com/questions/9160335/os-specific-instructions-in-cmake-how-to
```
if(MSVC OR MSYS OR MINGW)
    # for detecting Windows compilers
endif()
```

sloede:
> you may want to change your code to be conditional to WINDOWS

I do not think that's necessary. The proposed parameter settings should work for all Unix-like and Windows-like systems. In fact, I think it has been the default for a number of CMake releases now.

sloede:
@yspolyakov are there any updates on this issue? I'd be happy to update the PR such that it can be merged eventually (I don't think there is much to it anyways)

dsuponitskiy:
A new PR was open instead: https://github.com/openfheorg/openfhe-development/pull/1266

PR #639, opened 2024-01-04
Moved EvalSum, Rotation, Bootstrap and Automorphism keys to the same map

dsuponitskiy:


pascoec (approved):

src/pke/include/cryptocontext.h
```diff
@@ -2701,6 +2689,17 @@ class CryptoContextImpl : public Serializable {
     std::shared_ptr<std::map<usint, EvalKey<Element>>> EvalSumRowsKeyGen(const PrivateKey<Element> privateKey,
                                                                          const PublicKey<Element> publicKey = nullptr,
                                                                          usint rowSize = 0, usint subringDim = 0);
+
+    // [[deprecated(
```
yspolyakov, line 2693: Why is the deprecation logic commented out here and other places?

dsuponitskiy: It is the new code waiting to be un-commented after publicKey is removed. I can remove it

yspolyakov:
We should add the same logic for `MultiEvalAutomorphismKeyGen` and multiparty key gen methods based on it. Should we do it as part of a separate issue/PR?

src/pke/include/cryptocontext.h
yspolyakov: This is fine. It is OK to keep it as is.

yspolyakov (approved):

PR #640, opened 2024-01-08
Wrong conjugate function

mmazz:
Conjugate function has a wrong sign. I tested with CKKS and the output is the same vector.

In addition I add the standard std:: for the accumulate function.


yspolyakov:
Not needed. See https://openfhe.discourse.group/t/conjugate-implementation/1010 for more details

PR #642, opened 2024-01-10
Fix for slowdown at a large number of threads

pascoec:
- Fix for slowdown at a large number of threads
- Remove at() and [] access operators from DCRTPoly
- Force DCRTPoly::CRTInterpoate() to only use COEFFICIENT format
- Remove Precon variables from DropLastElementAndScale
- Updates to lib-benchmark poly-benchmark

### DIFF
--- a/src/core/include/lattice/hal/default/dcrtpoly-impl.h
+++ b/src/core/include/lattice/hal/default/dcrtpoly-impl.h
@@ -227,21 +227,11 @@ DCRTPolyImpl<VecType> DCRTPolyImpl<VecType>::CloneWithNoise(const DiscreteGaussi
 
 template <typename VecType>
 DCRTPolyImpl<VecType> DCRTPolyImpl<VecType>::CloneTowers(uint32_t startTower, uint32_t endTower) const {
-    std::vector<NativeInteger> m(endTower - startTower + 1);
-    std::vector<NativeInteger> r(endTower - startTower + 1);
-
-    for (uint32_t i = startTower; i <= endTower; i++) {
-        m[i - startTower] = m_params->GetParams()[i]->GetModulus();
-        r[i - startTower] = m_params->GetParams()[i]->GetRootOfUnity();
-    }
-
-    const auto co = m_params->GetCyclotomicOrder();
-    auto params   = std::make_shared<Params>(co, m, r);
+    auto cycorder = m_params->GetCyclotomicOrder();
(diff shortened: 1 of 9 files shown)


src/core/include/lattice/hal/default/dcrtpoly-impl.h
```diff
@@ -808,173 +778,83 @@ void DCRTPolyImpl<VecType>::ModReduce(const NativeInteger& t, const std::vector<
...
 template <typename VecType>
 typename DCRTPolyImpl<VecType>::PolyLargeType DCRTPolyImpl<VecType>::CRTInterpolate() const {
-    usint ringDimension = m_params->GetRingDimension();
-    usint nTowers       = m_vectors.size();
-
-    Integer bigModulus(m_params->GetModulus());  // qT
-
-    // this is the resulting vector of coefficients
-    VecType coefficients(ringDimension, bigModulus);
+    if (m_format == Format::EVALUATION)
```
dsuponitskiy, line 792: could it be replaced with "if (m_format != Format::COEFFICIENT)"?

src/core/include/lattice/hal/default/dcrtpoly-impl.h
```diff
@@ -808,173 +778,83 @@ void DCRTPolyImpl<VecType>::ModReduce(const NativeInteger& t, const std::vector<
...
-    VecType coefficients(ringDimension, bigModulus);
-
-    std::vector<Integer> multiplier(nTowers);
-    for (usint vi = 0; vi < nTowers; vi++) {
-        Integer qj(m_vectors[vi].GetModulus().ConvertToInt());
-        Integer divBy  = bigModulus / qj;
-        Integer modInv = divBy.ModInverse(qj).Mod(qj);
-        multiplier[vi] = divBy * modInv;
-    }
+    if (m_format == Format::EVALUATION)
```
dsuponitskiy, line 830: could it be replaced with "if (m_format != Format::COEFFICIENT)"?

src/core/include/lattice/hal/default/dcrtpoly-impl.h
pascoec: will update in PR for issue 458

src/core/include/lattice/hal/default/dcrtpoly-impl.h
pascoec: will update in PR for issue 458

PR #659, opened 2024-01-29
Fixed unittest failures when the library is linked with clang++ in th…

dsuponitskiy:
…e Debug mode and a number of sanitize flags (reported by Google)
Changes:
- made m_LWEscheme an object data member in BinFHEContext instead of being a shared_ptr as the uninitialized shared _ptr caused a crash. I can remove m_LWEscheme entirely, as LWEEncryptionScheme is a collection of function members only, and make all LWEEncryptionScheme's functions static to call them using the scope resolution operator. 
- updated code (comments only) according to https://github.com/openfheorg/openfhe-development/pull/632
- corrected code in src/binfhe/lib/lwe-pke.cpp (construction of shared_ptr)
- corrected and optimized a loop in src/core/include/math/discretegaussiangenerator-impl.h as it caused a crash 
- corrected code in src/pke/lib/scheme/ckksrns/ckksrns-leveledshe.cpp as it caused a crash


src/pke/lib/scheme/ckksrns/ckksrns-leveledshe.cpp
```diff
@@ -227,11 +227,32 @@ std::vector<DCRTPoly::Integer> LeveledSHECKKSRNS::GetElementForEvalAddOrSub(Cons
...
+    // -fsanitize=null
+    // -gz=zlib
+    // -fno-asynchronous-unwind-tables
+    // -fno-optimize-sibling-calls
+    // -fsplit-dwarf-inlining
+    // -gsimple-template-names
+    // -gsplit-dwarf
+    int32_t logApprox = 0;
+    const double res  = std::fabs(operand * scFactor);
+    if (res == 0) {
```
yspolyakov, line 246: Try if (res > 0) instead, just like below in the code

yspolyakov:
The change for m_LWEScheme should be reverted. Instead of using nullptr, initialize it using the default constructor.

pascoec (approved):

src/pke/lib/scheme/ckksrns/ckksrns-schemeswitching.cpp
```diff
@@ -637,17 +741,14 @@ std::vector<std::vector<NativeInteger>> ExtractLWEpacked(const Ciphertext<DCRTPo
...
-    auto& originalAVals = originalA.GetValues();
-    auto& originalBVals = originalB.GetValues();
-
-    extracted[1].insert(extracted[1].end(), &originalAVals[0], &originalAVals[N]);
-    extracted[0].insert(extracted[0].end(), &originalBVals[0], &originalBVals[N]);
+    // create 2 "begin" iterators to work with element values
+    const auto itA = std::vector<NativeInteger>::const_iterator(&originalA.GetValues()[0]);
+    const auto itB = std::vector<NativeInteger>::const_iterator(&originalB.GetValues()[0]);
+    std::vector<std::vector<NativeInteger>> extracted{std::vector<NativeInteger>(itB, itB + N),
+                                                      std::vector<NativeInteger>(itA, itA + N)};
```
pascoec, line 752: This is correct as written, but I would suggest changing L740-741 to
```
    auto originalA{(ct->GetElements()[0]).GetElementAtIndex(0)};
    auto originalB{(ct->GetElements()[1]).GetElementAtIndex(0)};
```
so that this becomes
```
    std::vector<std::vector<NativeInteger>> extracted{std::vector<NativeInteger>(itA, itA + N),
                                                      std::vector<NativeInteger>(itB, itB + N)};
```

pascoec (approved):

dsuponitskiy:
resolved after having talked with Carlo

PR #672, opened 2024-02-13
629 correct the use of compress in the scheme switching code

### LINKED ISSUE #629: Correct the use of Compress in the Scheme Switching code
Compress is called in two places in `ckkrns-schemeswitching.cpp`. The current version does not have the scale adjustment logic for FLEXIBLEAUTO and FLEXIBLEAUTOEXT. If the developers call it for this scaling mode, the approximation error will get increased to that of FIXED* modes. See the clarifications below.

When Compress is called for FLEXIBLEAUTO and FLEXIBLEAUTOEXT, it reduces the noise degree to 1 and then simply drops remaining RNS limbs (to the desired level). If all we do after this is decryption, this does not increase the approximation error for FLEXIBLEAUTO and FLEXIBLEAUTOEXT. Compress is written this way not to do extra work when Compress is called before normal decryption.

If we use it before any flavor of bootstrapping, we need to multiply the ciphertext by the appropriate adjustment factor that will set the scaling factor for the destination level. See https://github.com/openfheorg/openfhe-development/blob/v1.1.2/src/pke/lib/scheme/ckksrns/ckksrns-multiparty.cpp#L143 and the adjustment logic right after it for an example.

andreea-alexandru:
- Corrected the scaling factors for the FLEXIBLE modes
- Moved the initial scaling by the final scaling factor from the user's side to internal functions. This changes the examples and removes one argument from EvalCompareSwitchPrecompute.
- Clean-up

Closes #629

### DIFF
--- a/src/pke/unittest/utckksrns/UnitTestSchemeSwitch.cpp
+++ b/src/pke/unittest/utckksrns/UnitTestSchemeSwitch.cpp
@@ -320,18 +320,9 @@ protected:
             auto ccLWE          = cc->GetBinCCForSchemeSwitch();
             cc->EvalCKKStoFHEWKeyGen(keyPair, privateKeyFHEW);
 
-            const auto cryptoParams = std::dynamic_pointer_cast<CryptoParametersCKKSRNS>(cc->GetCryptoParameters());
-            ILDCRTParams<Element::Integer> elementParams = *(cryptoParams->GetElementParams());
-            auto paramsQ                                 = elementParams.GetParams();
-            auto modulus_CKKS_from                       = paramsQ[0]->GetModulus();
-            auto modulus_LWE                             = 1 << testData.logQ;
-            auto pLWE                                    = modulus_LWE / (2 * ccLWE->GetBeta().ConvertToInt());
-
-            double scFactor = cryptoParams->GetScalingFactorReal(0);
-            if (cryptoParams->GetScalingTechnique() == FLEXIBLEAUTOEXT)
-                scFactor = cryptoParams->GetScalingFactorReal(1);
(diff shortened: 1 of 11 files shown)


yspolyakov (approved):

src/pke/unittest/utckksrns/UnitTestSchemeSwitch.cpp
```diff
@@ -320,18 +320,9 @@ class UTCKKSRNS_SCHEMESWITCH : public ::testing::TestWithParam<TEST_CASE_UTCKKSR
...
-            auto pLWE                                    = modulus_LWE / (2 * ccLWE->GetBeta().ConvertToInt());
-
-            double scFactor = cryptoParams->GetScalingFactorReal(0);
-            if (cryptoParams->GetScalingTechnique() == FLEXIBLEAUTOEXT)
-                scFactor = cryptoParams->GetScalingFactorReal(1);
-            double scale = modulus_CKKS_from.ConvertToInt() / (scFactor * pLWE);
-
+            auto modulus_LWE = 1 << testData.logQ;
+            auto pLWE        = modulus_LWE / (2 * ccLWE->GetBeta().ConvertToInt());
+            double scale     = 1.0 / pLWE;
```
dsuponitskiy, line 325: is the check "if (pLWE != 0)" required here?

andreea-alexandru: No, it is not. modulus_LWE is obtained as a power of two, so it can never be 0. pLWE is modulus_LWE over something always positive.

dsuponitskiy (approved):
Looks good

PR #679, opened 2024-02-27
668 refactor openfhe exception class

dsuponitskiy:


src/core/include/utils/get-call-stack.h
```diff
@@ -31,17 +31,19 @@
...
 /**
  * @brief get_call_stack() is a function to get the call stack
  * @return a vector with call stack (demangled function names)
  */
 std::vector<std::string> get_call_stack();
+#else
+std::vector<std::string> get_call_stack() {
+    return std::vector<std::string>();
+}
 #endif
```
pascoec, line 48: remove this guard from header file and add specialized implementation to cpp file

dsuponitskiy: will do this for v1.2.0

pascoec (approved):

PR #685, opened 2024-03-01
Moved function's implementation from header to source file

dsuponitskiy:


jbates-duality (commented):
Straightforward move of code from header to implementation file.  Approving.

jbates-duality (approved):
Approved.

PR #724, opened 2024-03-27
Minor improvements for better Rust compatibility

### LINKED ISSUE #723: [Feature request] Improve interface of the keypair type
Is it possible to mark [keypair `good` method](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/include/key/keypair.h#L55) as a **const** one? Rust is very sensitive in terms of constantness. That would help us develop a better Rust trait in our [openfhe-rs wrapper](https://github.com/fairmath/openfhe-rs).

Also, it makes sense to choose **a better name** for the method, since the current name seems to be ambiguous. Maybe `is_allocated` would be a better option?

### LINKED ISSUE #726: [Feature request] Improve interface of the CryptoContextImpl type
As already mentioned in https://github.com/openfheorg/openfhe-development/issues/723 , Rust is very sensitive in terms of constantness. Correct `const` modifiers would help us develop a better Rust trait in our [openfhe-rs wrapper](https://github.com/fairmath/openfhe-rs).

It seems, that the following methods of the CryptoContextImpl class do not modify the instance. The implementation would be more accurate if they could be marked with the `const` modifier:
* [`KeyGen()`](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/include/cryptocontext.h#L1082) 
* [`SparseKeyGen()`](https://github.com/openfheorg/openfhe-development/blob/main/src/pke/include/cryptocontext.h#L1092)

nkaskov:
In this PR, we bring minor interface improvements in order to achieve better compatibility with the [Rust binding](https://github.com/fairmath/openfhe-rs) we are building.

The changes brought by this PR:
- Closes #723 
- Closes #726 
- We added additional `explicit` constuctor to the `CCParams` classes. This allows for the creation of instances of these classes using only the `Params` class.

### DIFF
--- a/src/pke/include/constants-fwd.h
+++ b/src/pke/include/constants-fwd.h
@@ -1,0 +1,136 @@
+//==================================================================================
+// BSD 2-Clause License
+//
+// Copyright (c) 2014-2022, NJIT, Duality Technologies Inc. and other contributors
+//
+// All rights reserved.
+//
+// Author TPOC: contact@openfhe.org
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this
+//    list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//==================================================================================
+
+#ifndef _CONSTANTS_FWD_H_
+#define _CONSTANTS_FWD_H_
+
+namespace lbcrypto {
+
+/**
+ * @brief Lists all features supported by public key encryption schemes
+ */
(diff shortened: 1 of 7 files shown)


yspolyakov:
@nkaskov Just want to confirm that this change is needed for Rust compatibility. I see changes in some example, implying that the user API has changed (unexpected behavior for v1.1.5), which may lead to compilation errors for existing projects based on OpenFHE.

Maybe it would be safer to add "is_allocated" w/o removing "good", so that the code would backwards compatible. In general, we only change API during major releases. Is this change required for the Rust bindings?

nkaskov:
> Maybe it would be safer to add "is_allocated" w/o removing "good", so that the code would backwards compatible. In general, we only change API during major releases. Is this change required for the Rust bindings?

Hi @yspolyakov! Thanks for your contribution. We decided to go with proposed solution and added `is_allocated()`, keeping `good()` method for the backward compatibility. The `good()` method has exactly the same interface as before, and `is_allocated()` has the interface required by the Rust binding.

src/pke/include/scheme/ckksrns/gen-cryptocontext-ckksrns-params.h
```diff
@@ -56,6 +56,7 @@ class CCParams<CryptoContextCKKSRNS> : public Params {
 public:
     CCParams() : Params(CKKSRNS_SCHEME) {}
     explicit CCParams(const std::vector<std::string>& vals) : Params(vals) {}
+    explicit CCParams(const Params& params) : Params(params) {}
```
dsuponitskiy, line 59: We do not need this constructor

src/pke/include/scheme/bgvrns/gen-cryptocontext-bgvrns-params.h
```diff
@@ -56,6 +56,7 @@ class CCParams<CryptoContextBGVRNS> : public Params {
 public:
     CCParams() : Params(BGVRNS_SCHEME) {}
     explicit CCParams(const std::vector<std::string>& vals) : Params(vals) {}
+    explicit CCParams(const Params& params) : Params(params) {}
```
dsuponitskiy, line 59: We do not need this constructor

src/pke/include/scheme/bfvrns/gen-cryptocontext-bfvrns-params.h
```diff
@@ -56,6 +56,7 @@ class CCParams<CryptoContextBFVRNS> : public Params {
 public:
     CCParams() : Params(BFVRNS_SCHEME) {}
     explicit CCParams(const std::vector<std::string>& vals) : Params(vals) {}
+    explicit CCParams(const Params& params) : Params(params) {}
```
dsuponitskiy, line 59: We do not need this constructor

src/pke/include/key/keypair.h
```diff
@@ -52,9 +52,13 @@ class KeyPair {
     explicit KeyPair(PublicKeyImpl<Element>* a = nullptr, PrivateKeyImpl<Element>* b = nullptr)
         : publicKey(a), secretKey(b) {}
 
-    bool good() {
+    bool is_allocated() const {
```
dsuponitskiy, line 55: as good() is already called in the code, i would make the existing function good() "const" and have is_allocated() call good().

src/pke/include/constants-fwd.h
```diff
@@ -0,0 +1,136 @@
+//==================================================================================
```
dsuponitskiy, line 1: I am aware about this situation, but didn't have time to fix it. it will be resolved differently, without creating a new file. just a few line code change

src/pke/include/key/keypair.h
nkaskov: Sure, no problem!

nkaskov: I will change this commit as you suggested.

dsuponitskiy:
a new PR has been submitted instead of this one. see it here: https://github.com/openfheorg/openfhe-development/pull/755

PR #735, opened 2024-04-11
721 add exceptions

### LINKED ISSUE #721: Add exception handling for ciphertext/ciphertext and ciphertext/plaintext operations when the number of RNS limbs is mismatched
This problem applies mostly to BFV (when Compress and/or encoding at a lower level are supported). CKKS and BGV handle this automatically for most of the modes.

I am suggesting a solution of updating all the *Core operations in `based-leveledshe.cpp` that perform actual operations (not redirecting to another operation) to check that the number of RNS limbs in both input objections is the same. I think the efficiency cost of this would be small.

andreea-alexandru:
- Added checks on the number of limbs in the core functions with two inputs (at deepest level).  @dsuponitskiy
- Fixed a plaintext generation with extra limbs which was throwing an exception.

Closes #721.

### DIFF
--- a/src/pke/include/schemebase/base-leveledshe.h
+++ b/src/pke/include/schemebase/base-leveledshe.h
@@ -784,6 +784,10 @@ protected:
     /////////////////////////////////////////
     // CORE OPERATIONS
     /////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


yspolyakov (approved):

dsuponitskiy (approved):
Looks good.

PR #737, opened 2024-04-15
Introduce VectorLWECipherText with cached validation flag

### LINKED ISSUE #1306: Add vector of LWE ciphertexts

arcturusannamalai:
- Introduce VectorLWECipherText with cached validation flag
   - this has potential to speedup the evaluation of binary operators etc.
- early exit when possible

### DIFF
--- a/src/binfhe/include/binfhe-base-scheme.h
+++ b/src/binfhe/include/binfhe-base-scheme.h
@@ -68,7 +68,7 @@ class VecLWECiphertext {
     bool validate() const;
 
     private:
-        mutable std::optional<bool> m_independent;
+        mutable std::optional<bool> m_independent = std::nullopt;
 };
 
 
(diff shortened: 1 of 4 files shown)


yspolyakov:
I suggest discussing this further. The idea of adding a vector of LWE ciphertexts is reasonable but it is worth discussing how to do it first.

arcturusannamalai:
@yspolyakov - thank you for the note; there are few use cases:
- the same argument sets are used to compute functions **f** and **g** and this type of caching can help
- lowering primitives from various transpilers maybe ensuring uniqueness of the cipher-text arguments to the OPenFHE primitives we can indicate uniqueness already

As for the implementation I can think of two strategies,
- Global register of vectors which are unique or not and update it (not my choice of implementation since its hard to do it threadsafe)
- Current implementation (need polish perhaps to propagate the solution through the OPenFHE API instead of wrap the arguments locally like I have done)  where each vector stores its validation flag within it; perhaps the validate() method could be private and set only at c-tor time or by invoking at the time of OPenFHE binary operator etc. 

Let me know what are your thoughts

PR #741, opened 2024-04-19
fixed a bug with mult depth = 0 for CKKS FLEXIBLEAUTO* modes

yspolyakov:
The wrong scaling factor was used for FLEXIBLEAUTO* modes at multiplicative depth = 0. It was using the first modulus size instead of the scaling factor. Now it is corrected.

### DIFF
--- a/src/pke/lib/scheme/ckksrns/ckksrns-cryptoparameters.cpp
+++ b/src/pke/lib/scheme/ckksrns/ckksrns-cryptoparameters.cpp
@@ -100,35 +100,19 @@ void CryptoParametersCKKSRNS::PrecomputeCRTTables(KeySwitchTechnique ksTech, Sca
         }
         else {
             m_scalingFactorsReal[0] = moduliQ[sizeQ - 1].ConvertToDouble();
-
-            if (extraBits == 0) {
-                for (uint32_t k = 1; k < sizeQ; k++) {
-                    double prevSF           = m_scalingFactorsReal[k - 1];
-                    m_scalingFactorsReal[k] = prevSF * prevSF / moduliQ[sizeQ - k].ConvertToDouble();
-                    double ratio            = m_scalingFactorsReal[k] / m_scalingFactorsReal[0];
-
-                    if (ratio <= 0.5 || ratio >= 2.0)
-                        OPENFHE_THROW(
-                            "CryptoParametersCKKSRNS::PrecomputeCRTTables "
-                            "- FLEXIBLEAUTO cannot support this "
-                            "number of levels in this parameter setting. Please use "
-                            "FIXEDMANUAL.");
-                }
-            }
-            else {
+            if (extraBits > 0)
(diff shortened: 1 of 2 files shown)


src/pke/lib/scheme/ckksrns/ckksrns-cryptoparameters.cpp
```diff
@@ -81,38 +81,54 @@ void CryptoParametersCKKSRNS::PrecomputeCRTTables(KeySwitchTechnique ksTech, Sca
...
-
-                if (ratio <= 0.5 || ratio >= 2.0)
-                    OPENFHE_THROW(
-                        "CryptoParametersCKKSRNS::PrecomputeCRTTables "
-                        "- FLEXIBLEAUTO cannot support this "
-                        "number of levels in this parameter setting. Please use "
-                        "FIXEDMANUAL.");
+            m_scalingFactorsReal[0] = moduliQ[sizeQ - 1].ConvertToDouble();
+
+            if (extraBits == 0) {
```
dsuponitskiy, line 104: this if statement may be re-written to avoid redundancies

yspolyakov: My biggest concern here is the readability of the code. The underlying logic is very complicated (and I made it more complicated with the fix). I believe you are referring to the prior branching for mult dept > 0. If you can find a way to do compress it and still preserve the clarity, please feel free to do it. I will need to review the revised branches to make sure the logic is preserved.

andreea-alexandru: It seems that based on the value of extraBits, we can set k0 = 1 or k0 = 2 and m_scalingFactorsReal[1], then execute outside the if the for loop starting from k0.

src/pke/lib/scheme/ckksrns/ckksrns-cryptoparameters.cpp
```diff
@@ -81,38 +81,38 @@ void CryptoParametersCKKSRNS::PrecomputeCRTTables(KeySwitchTechnique ksTech, Sca
...
-                double ratio            = m_scalingFactorsReal[k] / m_scalingFactorsReal[1];
-
+                double ratio            = m_scalingFactorsReal[k] / lastPresetFactor;
                 if (ratio <= 0.5 || ratio >= 2.0)
                     OPENFHE_THROW(
-                        "CryptoParametersCKKSRNS::PrecomputeCRTTables "
-                        "- FLEXIBLEAUTO cannot support this "
-                        "number of levels in this parameter setting. Please use "
-                        "FIXEDMANUAL.");
+                        "FLEXIBLEAUTO cannot support this number of levels in this parameter setting. Please use FIXEDMANUAL.");
```
yspolyakov, line 115: Change "Please use FIXEDMANUAL." to "Please use FIXEDMANUAL or FIXEDAUTO instead."

dsuponitskiy (approved):
Looks good

PR #742, opened 2024-04-19
blake2engine: Use .data() to get pointer to data and not .begin()

autoantwort:
`.begin()` could return an iterator object depending on the build context


autoantwort:
So you prefer one big PR with many unrelated changes instead of one PR per concern? 
Should I create one Issue for every PR / Concern or one Big one?

yspolyakov:
I suggest discussing the problem first. We want to understand the motivation before reviewing the PRs. Many PRs appeared very small. You can create an issue for each separate problem. A PR can afterwards be created for multiple issues.

autoantwort:
I have created #770

PR #752, opened 2024-04-29
Fix serialization issue for BFV scheme

g-s-m:
Fixing the serialization issue for BFV scheme


dsuponitskiy:
a new PR has been submitted instead of this one.

PR #755, opened 2024-04-29
Changes for better rust compatibility

dsuponitskiy:


dsuponitskiy:
Submitted instead of https://github.com/openfheorg/openfhe-development/pull/724

jbates-duality (approved):
This PR amounts to moving some constants [enums] from various files into a centralized file containing only that type of material.  Further, some methods that were const in nature have been expressly declared as such.  Straighforward.  Approved.

PR #763, opened 2024-05-03
Fix a bug in EvalSumCols

dsuponitskiy:


yspolyakov:
We need to discuss this further. I am closing the PR.

PR #768, opened 2024-05-15
756: Fixed EvalSumColsKeyGen() bug

dsuponitskiy:
1. Renamed the input parameter "rowSize" to numRows for EvalSumRows() or to numCols for EvalSumCols().
2. Replaced condition checks "SCHEME::CKKSRNS_SCHEME == schemeID" with "isCKKS(schemeID)".
3. Removed the unused input parameter "PublicKey<Element> publicKey" from protected/private functions
4. Replaced "usint" with "uint32_t"
5. Corrected some doxygen function documentations
6. Added a protected static helper function as a function members of Cryptocontext returning a pointer to a subset of the EvalAutomorphismKey map: GetPartialEvalAutomorphismKeyMapPtr(). This subset is built based on keyID and indexList
7. Added SerializeEvalAutomorphismKey() to serialize a subset of the EvalAutomorphismKey map given keyID and indexList.
8. Added DeserializeEvalAutomorphismKey() to deserialize the entire (or partial) serialized EvalAutomorphismKey map,  extract the keys by the given keyID and indexList only, add the extracted keys to the key map and discard the rest of the deserialized map of keys.
9. All "GenerateIndices..." functions return "std::unordered_set<uint32_t>" instead of "std::vector<usint>".
10. Added an additional internal variable to EvalSumRowsKeyGen() and to EvalSumColsKeyGen() in order to keep track of ALL required key indices and in order to returns ALL of the necessary keys.
11. EvalSumColsKeyGen() also generates (if they have not been generated yet) and returns all keys generated by EvalSumKeyGen() in addition to the keys EvalSumColsKeyGen() is supposed to generate by itself. For that a new function GenerateIndexListForEvalSum() was created.
12. Got rid of unnecessary variables and operations and slightly reorganized the code
13. Added unit tests


yspolyakov (approved):

src/pke/unittest/utckksrns/UnitTestCKKSrnsAutomorphism.cpp
```diff
@@ -353,6 +367,110 @@ class UTCKKSRNS_AUTOMORPHISM : public ::testing::TestWithParam<TEST_CASE_UTCKKSR
...
+    }
+
+    void UnitTest_EvalSumRows(const TEST_CASE_UTCKKSRNS_AUTOMORPHISM& testData,
+                              const std::string& failmsg = std::string()) {
+        try {
+            CryptoContext<Element> cc(UnitTestGenerateContext(testData.params));
+
+            KeyPair<Element> kp = cc->KeyGen();
+
+            std::vector<double> mat1 = {1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0};
```
yspolyakov, line 384: Change inputs to {1,2,3,4,5,7,8) and (8,1,2,3,4,5,6,7} (remove duplicates)

jbates-duality (approved):
Medium-sized patch with good PR commentary.

Code matches intent stated in commentary.  Many little improvements [isSCHEME(), nomenclature] other than just matching stated intent.  Good updates to doxygen.

Approving.

PR #788, opened 2024-05-31
Removed unnecessary const for the return value of GetRealPackedValue()

dsuponitskiy:


src/pke/include/encoding/ckkspackedencoding.h
```diff
@@ -153,7 +153,7 @@ class CKKSPackedEncoding : public PlaintextImpl {
         return value;
     }
 
-    const std::vector<double> GetRealPackedValue() const {
+    std::vector<double> GetRealPackedValue() const {
```
pascoec, line 156: add override?

pascoec (approved):

Issue #791, opened 2024-06-04
Review/update unit tests for FIXED_NOISE_HRA and NOISE_FLOODING_HRA


yspolyakov:
1. Make sure unit PRE tests are added in BGV, BFV, and CKKS for FIXED_NOISE_HRA.
2. Check unit tests for NOISE_FLOODING_HRA (BGV only). Add tests for 1, 4 and 5 hops. Use pre-hra-secure as an example.

pascoec:
adding fix for 4 hops

Issue #789, opened 2024-05-31
Fixes in the validator logic


yspolyakov:
1. scalingModSize should be enabled for the FIXEDMANUAL mode of BGV
3. statisticalSecurity and numAdversarialQueries should be enabled for BGV only for NOISE_FLOODING_HRA
4. statisticalSecurity and numAdversarialQueries should be enabled for CKKS only for NOISE_FLOODING_MULTIPARTY
5. NOISE_FLOODING_HRA should be enabled only for BGV and FIXEDMANUAL
6. PRENumHops should set to > 0 when NOISE_FLOODING_HRA is set; multiplicativeDepth should be set to 0 in this case

PR #792, opened 2024-06-05
Many changes in the HRA-secure PRE implementation based on BGV

yspolyakov:
The HRA-secure PRE is based on the algorithms proposed in https://eprint.iacr.org/2024/681

The following changes have been made:

1. The noise estimation logic has been updated to support HRA-secure PRE w/o any other homomorphic operations. The modulus size is now determined by the value of the noise flooding parameter.
2. Added an HRA-secure PRE example.
3. Renamed multiHopModSize with PRENumHops. So instead of asking for the bound on ciphertext modulus, OpenFHE now asks for the number of PRE hops to be supported in the case of HRA-secure PRE.
4. Fixed the noise switching estimate for BV keyswitching in BFV (added a multiplicative factor of 0.5).

Related issues: #789 and #791


dsuponitskiy (approved):
Looks good

andreea-alexandru (approved):
Look good.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
```diff
@@ -363,19 +373,75 @@ bool ParameterGenerationBGVRNS::ParamsGenBGVRNS(std::shared_ptr<CryptoParameters
         OPENFHE_THROW(s.str());
     }
 
+    uint32_t ringDimension = cyclOrder / 2;
+    InitializeFloodingDgg(cryptoParams, numHops, ringDimension);
+
     bool dcrtBitsSet = (dcrtBits == 0) ? false : true;
```
dsuponitskiy, line 379: Question: let's say dcrtBits = 0 and dcrtBitsSet=false. dcrtBits gets a non-zero value below if (scalTech == FIXEDMANUAL), but we still use dcrtBitsSet as "false" on line 472. I understand that dcrtBits changes its value for FIXEDMANUAL, but it is very confusing. I know we are supposed to work with the changes only, but I will get rid of dcrtBitsSet to avoid any possible future issues

yspolyakov: Now the dcrtBitsSet logic is captured by the validator. So it is OK to remove it since the validator changes have been made.

PR #794, opened 2024-06-06
Fixed tcmalloc crashes

dsuponitskiy:


jbates-duality (approved):
Straightforward pointer protection in demangle / exception code.  Approving.

PR #814, opened 2024-06-24
added special handling for BFV parameterization

yspolyakov:
Small fix for hybrid key switching parameterization in BFV


dsuponitskiy (approved):
Looks good

PR #837, opened 2024-07-18
Added new compiler support (up to g++-14 and clang++-18)

dsuponitskiy:


jbates-duality (approved):
Deprecate CLANG-12 and CLANG-13, add CLANG-18 and make it the default.

PR #847, opened 2024-08-17
Apple build has duplicate link commands

ai-mannamalai:
Remove duplicate commands for APPLE build


dsuponitskiy:
CMAKE_C_COMPILER_ID is used to identify the C compiler, not C++. Thus, I would keep it as we still use some C-language files

PR #850, opened 2024-08-20
Fix unnecessary data copies by correctly using std::move

lscat11:


dsuponitskiy:
This code change will be handled by [this PR](https://github.com/openfheorg/openfhe-development/pull/857)

PR #854, opened 2024-09-02
Fix FindSlotsToCoeffsRotationIndices() bug

lscat11:
It seems that `FindSlotsToCoeffsRotationIndices()` will return additional unused rotation keys.


andreea-alexandru:
The logic of going only up to levelBudget -1 is not correct; try running advanced-ckks-bootstrapping.cpp example and see that it fails because there are missing evaluation keys.

It should probably be levelBudget - flagRem. We'll test this logic and add it.

PR #857, opened 2024-09-03
Unnecessary data copy

dsuponitskiy:


dsuponitskiy:
The CI jobs succeeded

src/pke/lib/cryptocontext.cpp
```diff
@@ -788,30 +788,31 @@ std::unordered_map<uint32_t, DCRTPoly> CryptoContextImpl<DCRTPoly>::ShareKeys(co
...
 
                     NativeInteger powtemp(1);
                     for (size_t t = 1; t < threshold; t++) {
                         powtemp = powtemp.ModMul(i, modq_k);
 
-                        for (size_t d = 0; d < ring_dimension; d++) {
-                            powtempvec.at(d) = powtemp;
+                        NativeVector powtempvec(ring_dimension, modq_k);
+                        for (size_t i = 0; i < powtempvec.GetLength(); ++i) {
+                            // TODO (dsuponit): should we have a contructor to get a value for all m_data elements in NativeVector
```
dsuponitskiy, line 800: Change the comment to "... should we have a constructor to set all m_data elements in NativeVector to the same value?"

pascoec: The constructor already exists
NativeVectorT(usint length, const IntegerType& modulus, const IntegerType& val);

dsuponitskiy: That constructor sets all m_data elements to val.Mod(modulus), but we need to get them assigned to val

pascoec: It is supposed to be val.Mod(modulus). Equivalent to powtemp.Mod(modq_k).

pascoec (approved):

PR #859, opened 2024-09-06
Clear polynomials during CKKS decryption

yspolyakov:
Removes the integer values of polynomials so that the user could only access the floating-point values


dsuponitskiy (approved):
Looks good

PR #862, opened 2024-09-10
Updated the version of actions/checkout to v4 in 2 action.yml files

dsuponitskiy:


jbates-duality (commented):
Straightforward changes that only affect CI runs.  Approved.

PR #876, opened 2024-10-08
intt optimization

pascoec:
5% - 9% improvement for clang. No Change for gcc.

### DIFF
--- a/src/core/include/math/hal/intnat/transformnat-impl.h
+++ b/src/core/include/math/hal/intnat/transformnat-impl.h
@@ -350,7 +350,7 @@ void NumberTheoreticTransformNat<VecType>::ForwardTransformToBitReverseInPlace(c
         (*element)[i + 0] += omegaFactor - (omegaFactor >= (modulus - loVal) ? modulus : 0);
         if (omegaFactor > loVal)
             loVal += modulus;
-        (*element)[i + t] = loVal - omegaFactor;
+        (*element)[i + 1] = loVal - omegaFactor;
 #endif
     }
 }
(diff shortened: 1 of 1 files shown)


src/core/include/math/hal/intnat/transformnat-impl.h
```diff
@@ -350,7 +350,7 @@ void NumberTheoreticTransformNat<VecType>::ForwardTransformToBitReverseInPlace(c
         (*element)[i + 0] += omegaFactor - (omegaFactor >= (modulus - loVal) ? modulus : 0);
         if (omegaFactor > loVal)
             loVal += modulus;
-        (*element)[i + t] = loVal - omegaFactor;
+        (*element)[i + 1] = loVal - omegaFactor;
```
caesaretos, line 353: Why this change in Forward NTT? 
Please add a comment to explain the change

src/core/include/math/hal/intnat/transformnat-impl.h
```diff
@@ -551,6 +549,30 @@ void NumberTheoreticTransformNat<VecType>::InverseTransformFromBitReverseInPlace
...
+        (*element)[j1 + 0]  = loVal;
+        (*element)[j1 + j2] = omegaFactor;
+#else
+        (*element)[j1] += hiVal - (hiVal >= (modulus - loVal) ? modulus : 0);
+        auto omegaFactor = loVal + (hiVal > loVal ? modulus : 0) - hiVal;
+        omegaFactor.ModMulFastConstEq(omega1Inv, modulus, preconOmega1Inv);
+        (*element)[j1 + j2] = omegaFactor;
+#endif
+    }
+    for (uint32_t i = 0; i < j2; ++i)
```
caesaretos, line 574: Why do we need more than 1 multiply with N inverse?
The optimization I proposed should absorb all the multiply by N inverse, and we only need to multiply (*element)[0] by N inverse.

src/core/include/math/hal/intnat/transformnat-impl.h
pascoec: It's a silent bug. Should be constant 1. Previously worked because t is always 1 by the time this code is executed.

caesaretos (changes requested):
Please add more comments to explain the implementation of the optimization. It is hard to understand the code this way.

src/core/include/math/hal/intnat/transformnat-impl.h
pascoec: The optimization you originally proposed, which incorporated (N-1) inverse multiplies incorporated into the twiddles, would require ~2x twiddle memory, wouldn't be backwards compatible, and would require extra if statements to select appropriate twiddle data making the optimized transform run slower than the original. This code only incorporates the inverse into bitreversed(omega1), which is multiplied by the last N/2 elements in the final intt stage. This loop is added to multiply the remaining N/2 elements by the inverse. I would have added this multiply into the final intt stage as well, but benchmarking showed it was faster to just have an extra loop at the end.

yspolyakov:
I am concerned about the readability of the NTT code. It is becoming hard to understand.

yspolyakov (approved):

PR #878, opened 2024-10-16
852 support prngs v2

dsuponitskiy:
Add support for external PRNG:
- an external PRNG library is available [here](https://github.com/openfheorg/openfhe-prng-blake2)
- src/core/examples/external-prng.cpp was added to test external libraries


src/core/include/utils/prng/prng.h
```diff
@@ -0,0 +1,110 @@
...
+
+protected:
+    PRNG() = default;
+    PRNG(const std::array<PRNG::result_type, MAX_SEED_GENS> &seed, PRNG::result_type counter)
+        : m_counter(counter), m_seed(seed) {}
+
+    PRNG(const PRNG& other) {
+        m_counter = other.m_counter;
+        m_seed = other.m_seed;
+    }
```
jbates-duality, line 99: This copy constructor should be considered not-harmless.  To be concise, this may end up tolerating the [perhaps accidental] existence of multiple PRNGs outputting the same stream.  Very, very bad.

src/core/include/utils/prng/prng.h
```diff
@@ -0,0 +1,110 @@
...
+
+    // counter used as input to the hash function; gets incremented after each call
+    // TODO (dsuponit): do we need uint64_t as the m_counter type as we have
+    // PRNG::result_type for all other "counters"
+    uint64_t m_counter = 0;
+
+    // the seed for the hash function
+    std::array<PRNG::result_type, PRNG::MAX_SEED_GENS> m_seed{};
```
jbates-duality, line 107: RESCIND: m_seed is the bulk of the context for the chained hash. Nevermind.

IGNORE: Keeping the seed around just makes a replay attack that much easier.  Disrecommend.  Suggest not even having this member element, and using the passed seed directly in ctor, erasing it before returning.

src/core/include/utils/prng/prng.h
```diff
@@ -0,0 +1,110 @@
...
+      return std::numeric_limits<result_type>::max();
+    }
+
+    virtual result_type operator()() = 0;
+    virtual ~PRNG() = default;
+
+protected:
+    PRNG() = default;
+    PRNG(const std::array<PRNG::result_type, MAX_SEED_GENS> &seed, PRNG::result_type counter)
+        : m_counter(counter), m_seed(seed) {}
```
jbates-duality, line 94: RESCIND: m_seed is the bulk of the context for the chained hash.  Nevermind.

IGNORE: Suggest immediately calling operator(), and erasing the seed, instead of this being an empty inline.  Even better see comment on m_seed, below.  We should not be keeping copies of seed strewn about.

src/core/lib/math/distributiongenerator.cpp
```diff
@@ -35,12 +35,142 @@
...
+} 
+
+PRNG& PseudoRandomNumberGenerator::GetPRNG() {
+    // initialization of PRNGs
+    if (m_prng == nullptr) {
+#pragma omp critical
+        {
+            if (!genPRNGEngine)
+                InitPRNGEngine();
+
```
jbates-duality, line 93: Should there be a check here that genPRNGEngine is no longer nullptr?

I realize this check is perfomed in the called code.  I also realize that it is declared static, so not able to be overridden.

This may be a no-op comment.

src/core/lib/math/distributiongenerator.cpp
```diff
@@ -35,12 +35,142 @@
...
+                    }
+                    rdGenPassed = true;
+                }
+                catch (std::exception& e) {
+                }
+            }
+            for (uint32_t i = 0; i < PRNG::MAX_SEED_GENS; ++i)
+                seed[i] += rdseed[i];
+  #endif  // FIXED_SEED
+            m_prng = std::shared_ptr<PRNG>(genPRNGEngine(seed, 0));
```
jbates-duality, line 168: This copy of the seed should be erased, right here.

src/core/include/utils/prng/prng.h
jbates-duality, line 107: Suggest a typedef for the seed, which promotes calling code to use it, which might avoid an unintended copy when converting from a non-matching data type through an unintended std::array<> ctor call.

src/core/lib/math/distributiongenerator.cpp
```diff
@@ -35,12 +35,142 @@
...
+            if (sizeof(size_t) == 8)
+                initKey[2] = (std::hash<std::thread::id>{}(std::this_thread::get_id()) >> 32);
+    #endif
+            // heap variable; we are going to use the least 32 bits of its memory
+            // location as the counter. This will increase the entropy of the PRNG sample
+            void* mem        = malloc(1);
+            uint32_t counter = reinterpret_cast<long long>(mem);  // NOLINT
+            free(mem);
+
+            std::uniform_int_distribution<uint32_t> distribution(0);
```
jbates-duality, line 135: Does std::uniform_int_distribution actually do anything here?  We expect the output of the PRNG to already be uniformly-distributed.  This has the appearance of a complicating no-op.  AmIRight?

src/core/lib/math/distributiongenerator.cpp
```diff
@@ -35,12 +35,142 @@
...
+                        rds = distribution(genR);
+                    }
+                    rdGenPassed = true;
+                }
+                catch (std::exception& e) {
+                }
+            }
+            for (uint32_t i = 0; i < PRNG::MAX_SEED_GENS; ++i)
+                seed[i] += rdseed[i];
+  #endif  // FIXED_SEED
```
jbates-duality, line 167: rdseed should be erased here, as this contains the majority of the seed entropy.

src/core/include/utils/prng/blake2engine.h
```diff
@@ -29,143 +29,77 @@
...
+    // The vector that stores random samples generated using the hash function
+    std::array<PRNG::result_type, PRNG::PRNG_BUFFER_SIZE> m_buffer{};
 
-  // The vector that stores random samples generated using the hash function
-  std::array<result_type, PRNG_BUFFER_SIZE> m_buffer{};
-
-  // Index in m_buffer corresponding to the current PRNG sample
-  uint16_t m_bufferIndex = 0;
+    // Index in m_buffer corresponding to the current PRNG sample
+    uint16_t m_bufferIndex = 0;
```
jbates-duality, line 89: size_t instead of uint16_t?

jbates-duality (changes requested):
Various comments regarding copies of seeds, dangerous copy constructors, possibly-erroneous unit test, #ifdef'ing for just linux.

CMakeLists.txt
```diff
@@ -185,6 +185,12 @@ if(WITH_COVTEST)
     set( COVDIR ${BUILDDIR}coverage/)
 endif()
 
+if(UNIX AND NOT APPLE AND "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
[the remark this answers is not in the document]
dsuponitskiy: right, i have this already prepared for the commit with corrections.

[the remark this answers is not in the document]
dsuponitskiy: My build with clang++ and the new compiler option "-ldl" failed on Ubuntu. even though clang uses the same option "-ldl". porting this code to macOS and MinGW may take time.

src/core/include/utils/prng/blake2engine.h
dsuponitskiy: missed this one. will correct

src/core/include/utils/prng/prng.h
dsuponitskiy: m_seed will go, if the copy constructor goes. will discuss this

src/core/include/utils/prng/prng.h
```diff
@@ -0,0 +1,110 @@
...
+
+    PRNG(const PRNG& other) {
+        m_counter = other.m_counter;
+        m_seed = other.m_seed;
+    }
+
+    // counter used as input to the hash function; gets incremented after each call
+    // TODO (dsuponit): do we need uint64_t as the m_counter type as we have
+    // PRNG::result_type for all other "counters"
+    uint64_t m_counter = 0;
```
jbates-duality, line 104: This is a 64-bit integer.  It is initialized with a 32-bit integer derived from a pointer address in distributiongenerator.h.

Shall we initialize this with a full 64 bits, instead of truncating to 32?

src/core/lib/math/distributiongenerator.cpp
```diff
@@ -35,12 +35,142 @@
...
+            // On a 64-bit machine, the thread id is 64 bits long
+            // skip on 32-bit arm architectures
+    #if !defined(__arm__) && !defined(__EMSCRIPTEN__)
+            if (sizeof(size_t) == 8)
+                initKey[2] = (std::hash<std::thread::id>{}(std::this_thread::get_id()) >> 32);
+    #endif
+            // heap variable; we are going to use the least 32 bits of its memory
+            // location as the counter. This will increase the entropy of the PRNG sample
+            void* mem        = malloc(1);
+            uint32_t counter = reinterpret_cast<long long>(mem);  // NOLINT
```
jbates-duality, line 132: We are truncating the pointer to 32-bits, here.  Internally to the Blake2Engine, m_counter is 64-bits.  Shall we alter the API to not truncate to 32?

src/core/unittest/UnitTestTransform.cpp
```diff
@@ -280,11 +280,9 @@ void CRT_CHECK_big_ring(const std::string& msg) {
     ChineseRemainderTransformArb<V>().SetCylotomicPolynomial(cycloPoly, modulus);
 
     V input(n, modulus);
-    PRNG gen(1);
```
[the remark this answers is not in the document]
jbates-duality: Per Yuriy, no need for a fixed sequence for this unit test.

jbates-duality (approved):

jbates-duality (commented):
Seed is still not a typedef.

yspolyakov (approved):

PR #880, opened 2024-10-21
Added unit tests for larger plaintext moduli in BFV

yspolyakov:
Added unit tests for 46-bit plaintext moduli


dsuponitskiy (approved):
Looks good

PR #883, opened 2024-10-22
Fixed the estimation logic for hybrid key switching

yspolyakov:
1. Incremented the estimated modulus for BGV and CKKS by 1 to account for RNS moduli above the desired bit width
2. Incremented the maximum digit size in hybrid key switching for BGV and CKKS by 1 to account for RNS moduli above the desired bit width

### DIFF
--- a/src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
+++ b/src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
@@ -444,10 +444,13 @@ bool ParameterGenerationBGVRNS::ParamsGenBGVRNS(std::shared_ptr<CryptoParameters
     if (multipartyMode == NOISE_FLOODING_MULTIPARTY)
         qBound += cryptoParamsBGVRNS->EstimateMultipartyFloodingLogQ();
 
+    // we add an extra bit to account for the special logic of selecting the RNS moduli in BGV
+    qBound++;
+
     uint32_t auxTowers = 0;
     if (ksTech == HYBRID) {
         auto hybridKSInfo =
-            CryptoParametersRNS::EstimateLogP(numPartQ, firstModSize, dcrtBits, extraModSize, numPrimes, auxBits);
+            CryptoParametersRNS::EstimateLogP(numPartQ, firstModSize, dcrtBits, extraModSize, numPrimes, auxBits, true);
         qBound += std::get<0>(hybridKSInfo);
         auxTowers = std::get<1>(hybridKSInfo);
     }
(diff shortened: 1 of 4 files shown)


src/pke/lib/schemerns/rns-cryptoparameters.cpp
```diff
@@ -426,6 +426,10 @@ std::pair<double, uint32_t> CryptoParametersRNS::EstimateLogP(uint32_t numPartQ,
             maxBits = bits;
     }
 
+    // we add an extra bit to account for for the special moduli selection logic in BGV and CKKS
+    if (addOne)
+        maxBits++;
```
andreea-alexandru, line 431: This adds at most one extra P limb by pushing the ceil of maxBits / auxBits to the next integer. This is probably enough for typical applications. But what if there is something like multiplicative depth 200, and there are more than 60 bits added because the qi's have dcrtBits+1 instead of dcrtBits bits? Shouldn't the number of buffer bits added depend on the number of qi's?

yspolyakov: I think this scenario is unrealistic because this would require more than 60 digits in HKS (the key would be huge). If someone uses 200 levels, they will probably use 3 digits to keep the evaluation keys comparable to ciphertexts in size.

yspolyakov: Actually I don't think this is possible. Max bits is found over all digits. Basically we find maxbits over all digits and then we add 1 to it.

src/pke/lib/scheme/bgvrns/bgvrns-parametergeneration.cpp
```diff
@@ -444,10 +444,13 @@ bool ParameterGenerationBGVRNS::ParamsGenBGVRNS(std::shared_ptr<CryptoParameters
     if (multipartyMode == NOISE_FLOODING_MULTIPARTY)
         qBound += cryptoParamsBGVRNS->EstimateMultipartyFloodingLogQ();
 
+    // we add an extra bit to account for the special logic of selecting the RNS moduli in BGV
+    qBound++;
```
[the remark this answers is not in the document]
yspolyakov: Removed the branch

src/pke/lib/schemerns/rns-cryptoparameters.cpp
andreea-alexandru: Makes sense.

andreea-alexandru (approved):
The logic looks good.

dsuponitskiy (approved):
Looks good

PR #888, opened 2024-10-23
Remove the extra keys in homomorphic decode

andreea-alexandru:
In the SlotsToCoefficients implementation via collapsed FFT, when the levels cannot be collapsed exactly in the given budget, there were some extra rotation keys being generated. Those extra keys are now removed.


dsuponitskiy (approved):
Looks good

PR #890, opened 2024-10-23
Update quickstart.rst

dsuponitskiy:
link to simple-real-numbers.cpp was broken


jbates-duality (approved):
Straightforward change to point to correct file.

PR #912, opened 2024-11-23
Added a validation for the cryptocontext params: digitSize should be less than 32 if keySwitchTechnique == BV (BGV/BFV)

dsuponitskiy:
Validate digitSize for BFV/BGV:
it should not be greater than (MAX_MODULUS_SIZE/2) if keySwitchTechnique is set to "BV"


yspolyakov:
Please also add the same check for CKKS

yspolyakov (approved):

PR #916, opened 2024-11-27
Moved schemeswitching-data-serializer.h from pke/include/scheme/ckksrns to pke/unittest as it is a collection of special utilities

dsuponitskiy:


andreea-alexandru (commented):
Looks good, but please add a paragraph in SCHEME_SWITCHING_CAPABILITY.md to say that an example for serializing scheme switching is given in scheme-switching-serial.cpp and users should also update src/pke/unittest/utils/schemeswitching-data-serializer.h if they want to specify other filenames.

andreea-alexandru (approved):
Good to go.

PR #918, opened 2024-12-04
Prevent integer overflow in FindAutomorphismIndex2nComplex()

dsuponitskiy:
1. Changed the type of **g0 and g from int32_t to uint64_t**:
    - to prevent integer overflow if the result of (g*g0) is too high
    - NativeInteger().ModInverse() always returns positive number, so the type can be unsigned
2. Replaced "if (i == int32_t(m - 1))" with "if (i == (static_cast<int32_t>(m) - 1))" as the subtraction should be performed on int32_t to avoid unexpected behavior. 
3. Replaced the C-style cast with the C++ one

### DIFF
--- a/src/core/lib/math/nbtheory2.cpp
+++ b/src/core/lib/math/nbtheory2.cpp
@@ -247,6 +247,8 @@ uint32_t FindAutomorphismIndex2nComplex(int32_t i, uint32_t m) {
     else if (i == (static_cast<int32_t>(m) - 1)) { // could be true if (i > 0)
         return static_cast<uint32_t>(i);
     }
+    if (!IsPowerOfTwo(m))
+        OPENFHE_THROW("m should be a power of two.");
 
     // conjugation automorphism
     // generator. the usage of uint64_t prevents occasional integer overflow if the result of (g*g0) is too high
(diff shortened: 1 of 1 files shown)


src/core/lib/math/nbtheory2.cpp
```diff
@@ -247,29 +244,19 @@ uint32_t FindAutomorphismIndex2nComplex(int32_t i, uint32_t m) {
...
-            g = (g * g0) % m;
-        }
-        return uint32_t(g);
+    // conjugation automorphism
+    // generator. the usage of uint64_t prevents occasional integer overflow if the result of (g*g0) is too high
+    const uint64_t g0   = (i < 0) ? NativeInteger(5).ModInverse(m).ConvertToInt() : 5;
+    uint64_t g          = g0;
+    uint32_t i_unsigned = static_cast<uint32_t>(std::abs(i));
+    for (size_t j = 1; j < i_unsigned; j++) {
+        g = (g * g0) % m;
```
yspolyakov, line 257: m is always a power of two. So we can use bit shifting instead of %. Just in case, check that m is a power of two at the beginning.

yspolyakov (approved):

PR #947, opened 2025-03-18
Interactive 2party bootstrapping

yspolyakov:
See Appendix D of https://eprint.iacr.org/2023/1203 for details.

* Initially enabled it only for CKKS. However, later this implementation can be extended to BGV and BFV. So the functions were added with this in mind.
* Added IntBootDecrypt, IntBootAdd, and IntBootEncrypt to the parent RNS scheme (MultipartyRNS).
* Added IntBootAdjustScale (special scaling for CKKS) to MultipartyCKKSRNS.
* Initially only one extra tower will be used for bootstrapping. Later we will make it a configurable parameter. To add the tower, the multiplicative depth needs to be increased by 1.


yspolyakov:
Notes for @dsuponitskiy 

1. Convert unit-test-interactive-bootstrapping.cpp to unit tests. Include FIXEDMANUAL, FIXEDAUTO, and FIXEDAUTO. Use 1024 as the ring dimension.
2. Add unit tests for all scenarios of interactive-bootstrapping.cpp.
3. Review the code with the main logic.

yspolyakov (commented):
I still need to review clean up interactive-bootstrapping.cpp

dsuponitskiy (approved):

PR #958, opened 2025-04-03
Cleartext version of EvalChebyshevFunction

shaih:
Added a few lines to the documentation of EvalChebyshevCoefficients, and also added a function EvalChebyshevFunctionPtxt that computes the same thing as CryptoContext->EvalChebyshevFunction but on cleartext vector<double>

### DIFF
--- a/src/core/include/math/chebyshev.h
+++ b/src/core/include/math/chebyshev.h
@@ -50,9 +50,11 @@ namespace lbcrypto {
 
 /**
  * Method for calculating Chebyshev coefficients for an input function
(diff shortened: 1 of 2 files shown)


yspolyakov:
Thank you for the contribution @shaih. Could you create a local branch in openfhe-development from `dev` (not `main`) and resubmit the PR? Using a local branch will allow us to run the CI against the PR. In this case, I can also make ``cosmetic'' edits (like revising the text for c[0]/2 to explain why division by 2 is done) without asking to push additional commits.

PR #1012, opened 2025-07-05
Disabled the Xallocator test

dsuponitskiy:


dsuponitskiy:
The Xallocator code is immature (experimental) and requires manual tuning. It is not meant for general usage. Because of that, we don’t use Xallocator or run unit tests with it. We previously did disable all Xallocator unit tests, but it seems that some code changes enabled them again.

PR #1030, opened 2025-07-31
Add sparse encapsulation for CKKS-based bootstrapping

### LINKED ISSUE #1001: Add sparse-secret encapsulation for CKKS (functional) bootstrapping
Implement the optimization from https://eprint.iacr.org/2022/024.

Right before `ModRaise`, switch from a denser key to a sparser key, then switch back to a denser key after `ModRaise`.

- [x] Add a new secret key distribution `SPARSE_ENCAPSULATED`, specifically for bootstrapping.
- [x] Generate two switching keys in `EvalBootstrapKeyGen` and the appropriate key switching calls to the code.
   - [x] For the second key generation and switching, regular key switching mechanism can be used.
    - [x] For the first key switching, we need to go from $q$ to $qp$, where $p$ can be the first auxiliary modulus in $P$. Then we need to rescale by $p$.
       * For the sparsest secret, use $p=60$ and the Hamming weight of 32. 
       * Add KeySwitchGenSparse and KeySwitchSparse methods specifically the first key switching inside the FHE private space.
- [x] Generate new Chebyshev series coefficients for regular CKKS bootstrapping (for $K=16$).
- [x] Generate new Chebyshev series coefficients for functional CKKS bootstrapping (for $K=16$).
- [x] Add unit tests for regular CKKS bootstrapping.
- [x] Throw exception in the following cases:
       * SPARSE_ENCAPSULATED will not work for COMPOSITESCALING* modes (yet). 
       * SPARSE_ENCAPSULATED will only work for CKKS. 
- [x] Add the SPARSE_ENCAPSULATED mode and related unit tests for functional bootstrapping.

yspolyakov:
See issue #1001 for details,


pascoec (approved):

src/pke/examples/test-sparse-encapsulation.cpp
```diff
@@ -0,0 +1,112 @@
+//==================================================================================
```
yspolyakov: I suggest converting this example to a unit test

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
```diff
@@ -2645,74 +2682,59 @@ Plaintext FHECKKSRNS::MakeAuxPlaintext(const CryptoContextImpl<DCRTPoly>& cc, co
...
     result->SetScalingFactor(result->GetScalingFactor() * plaintext->GetScalingFactor());
     return result;
 }
 
 void FHECKKSRNS::EvalAddExtInPlace(Ciphertext<DCRTPoly>& ciphertext1, ConstCiphertext<DCRTPoly> ciphertext2) const {
-    std::vector<DCRTPoly>& cv1       = ciphertext1->GetElements();
-    const std::vector<DCRTPoly>& cv2 = ciphertext2->GetElements();
-    for (size_t i = 0; i < cv1.size(); ++i)
+    auto& cv1  = ciphertext1->GetElements();
+    auto& cv2  = ciphertext2->GetElements();
```
yspolyakov, line 2698: Is it correct to drop `const` here? The intent was to say that cv2 will not be changed.

yspolyakov:
A general comment. I see many large-parameter unit tests. These probably take a while to run. In Google tests, one can set up  long tests and run those only when desired. I believe we even have this somewhere in OpenFHE (or had it before). I would not run many large-ring-dimension tests by default.

pascoec:
> A general comment. I see many large-parameter unit tests. These probably take a while to run. In Google tests, one can set up long tests and run those only when desired. I believe we even have this somewhere in OpenFHE (or had it before). I would not run many large-ring-dimension tests by default.

Are you referring to the Functional Bootstrapping unit tests? Two things:
(1) Those large-parameter values are used only when `BENCH` is defined. When `Bench` is undefined, the much smaller default values are used.
(2) The current unit test setup is temporary. One of the first things I plan to do in the Functional Bootstrapping refactor is create another set of fewer tests that are run instead of the current set with small-parameter values explicitly listed in the test vectors, allowing me to remove the current `#ifndef BENCH` logic that overrides the large-parameter values with the defaults. It will be a lot more straight forward.

src/pke/examples/test-sparse-encapsulation.cpp
pascoec: where would you like me to add it?

yspolyakov:
> > A general comment. I see many large-parameter unit tests. These probably take a while to run. In Google tests, one can set up long tests and run those only when desired. I believe we even have this somewhere in OpenFHE (or had it before). I would not run many large-ring-dimension tests by default.
> 
> Are you referring to the Functional Bootstrapping unit tests? Two things: (1) Those large-parameter values are used only when `BENCH` is defined. When `Bench` is undefined, the much smaller default values are used. (2) The current unit test setup is temporary. One of the first things I plan to do in the Functional Bootstrapping refactor is create another set of fewer tests that are run instead of the current set with small-parameter values explicitly listed in the test vectors, allowing me to remove the current `#ifndef BENCH` logic that overrides the large-parameter values with the defaults. It will be a lot more straight forward.

Sounds good. Thanks.

src/pke/examples/test-sparse-encapsulation.cpp
yspolyakov: UnitTestBootstrap.cpp under ckks unit tests. It is only needed for bootstrapping.

pascoec: will do

dsuponitskiy (approved):

PR #1052, opened 2025-09-09
EvalFastRotation: add an overload that computes cyclotomic order

### LINKED ISSUE #1051: EvalFastRotation should not need the cyclotomic order as a function argument
Cf. https://openfhe.discourse.group/t/why-does-evalfastrotation-require-the-cyclotomic-order-as-input/2152

`EvalFastRotation` (https://github.com/openfheorg/openfhe-development/blob/aa8a86e1143f1e47d4354bdd757080e903ba5875/src/pke/include/cryptocontext.h#L2385-L2394) has an argument `m` that is the cyclotomic order of the ring. This is always going to be computed as N/2 from the crypto context anyway (`cc->GetRingDimension() * 2` shows up in many examples/tests), so it should probably be removed or set to a default value so that callers do not need to specify it. Probably the following overload suffices

````cpp
  Ciphertext<Element> EvalFastRotation(ConstCiphertext<Element>& ciphertext, const uint32_t index,
                                       const std::shared_ptr<std::vector<Element>> digits) const {
      return EvalFastRotation(ciphertext, index, cc->GetRingDimension() * 2, digits);
  }
```

j2kun:
Fixes https://github.com/openfheorg/openfhe-development/issues/1051

### DIFF
--- a/src/pke/include/cryptocontext.h
+++ b/src/pke/include/cryptocontext.h
@@ -2394,6 +2394,39 @@ public:
     }
 
     /**
+    * @brief Implements the automorphism and key switching step of hoisted automorphisms.
+    *
+    * Please refer to Section 5 of Halevi and Shoup, "Faster Homomorphic linear transformations in HELib."
+    * for more details, link: https://eprint.iacr.org/2018/244.
+    *
+    * Generally, automorphisms are performed with three steps:
+    * (1) the automorphism is applied on the ciphertext
+    * (2) the automorphed values are decomposed into digits
+    * (3) key switching is applied to make it possible to further compute on the ciphertext.
+    *
+    * Hoisted automorphisms is a technique that performs the digit decomposition for the original ciphertext first,
+    * and then performs the automorphism and the key switching on the decomposed digits. The benefit of this is that the
+    * digit decomposition is independent of the automorphism rotation index, so  it can be reused for
+    * multiple different indices. This can greatly improve performance when we have to compute many automorphisms
+    * on the same ciphertext. This routinely happens when we do permutations (EvalPermute).
+    *
+    * This method assumes that all required rotation keys exist. This may not be true if we are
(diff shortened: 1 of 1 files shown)


src/pke/include/cryptocontext.h
```diff
@@ -2393,6 +2393,39 @@ class CryptoContextImpl : public Serializable {
...
+    * The cyclotomic order is computed from the CryptoContext.
+    *
+    * @param ciphertext  Input ciphertext.
+    * @param index       Rotation index (positive for left, negative for right).
+    * @param digits      Precomputed rotation data (the digit decomposition created by EvalFastRotationPrecompute).
+    * @return Rotated ciphertext.
+    */
+    Ciphertext<Element> EvalFastRotation(ConstCiphertext<Element>& ciphertext, const uint32_t index,
+                                         const std::shared_ptr<std::vector<Element>> digits) const {
+        return EvalFastRotation(ciphertext, index, cc->GetRingDimension() * 2, digits);
```
yspolyakov, line 2426: Why is 'cc` used here? I do not see it passed as an argument. Do you mean something like `GetRingDimension()` (i.e., without `cc`)?

yspolyakov:
@j2kun I suggest creating feature branches directly in the `openfhe-development` repo so we could run CI against them. Could you submit the revised code this way?

j2kun:
> @j2kun I suggest creating feature branches directly in the `openfhe-development` repo so we could run CI against them. Could you submit the revised code this way?

I will do this, though I suspect you can run CI against PRs by adding `dev` to the list here: https://github.com/openfheorg/openfhe-development/blob/aa8a86e1143f1e47d4354bdd757080e903ba5875/.github/workflows/pull-request.yml#L6

src/pke/include/cryptocontext.h
j2kun: Fixed

PR #1054, opened 2025-09-26
930 remove sudo from uninstall_openfhe.sh

### LINKED ISSUE #930: Remove "sudo" from scripts/uninstall_openfhe.sh
"sudo" is only required by uninstall if you install OpenFHE in /usr/local. but in this case "sudo" should be provided explicitly instead of being used in the uninstall script. Remove "sudo" from scripts/uninstall_openfhe.sh

dsuponitskiy:


jbates-duality (approved):
Absolutely straightforward removal of sudo from script, such that user calls the script with sudo on command-line instead.

PR #1067, opened 2025-10-15
Make cushion in BFV HPSPOVERQLEVELED mode larger

### LINKED ISSUE #1066: BFV HPSPOVERQLEVELED increased noise growth after #1007
After #1007, the noise estimate is more conservative for BFV.

In HPSPOVERQLEVELED mode, the `FindLevelsToDrop` function in `bfvrns-leveledshe.cpp` need to precisely estimate the noise (see https://eprint.iacr.org/2021/204, section 3.2, leveled bfv multiplication), to do so, OpenFHE used an empirical _cushion_ as the expansion factor based estimate is often much larger than the real noise. Namely, `estimate - cushion = real noise`

However, as the estimate is larger now, the cushion should be larger. Otherwise, if we overestimate the noise, and the modulus switching error will become dominant, violating eq. 12 in the above paper; after scaling back from Ql to Q, the modulus switching error will become larger than the original noise, and experimentally we can observe an increase in the overall noise growth. I found that after depth-20 multiplication, the noise is 30bit larger than previously. 

An interpretation of #1007 is that for each BFV mulplicativeDepth we will have one bit more estimate. @yspolyakov suggested changing the cushion formula from `2 * mulDepth` to `4 * mulDepth`, and it did make the noise in my experiment to behave similarly as before.

ZenithalHourlyRate:
Fixes #1066

Implements the fix suggested below:

> An interpretation of https://github.com/openfheorg/openfhe-development/pull/1007 is that for each BFV mulplicativeDepth we will have one bit more estimate. @yspolyakov suggested changing the cushion formula from 2 * mulDepth to 4 * mulDepth, and it did make the noise in my experiment to behave similarly as before.

### DIFF
--- a/src/pke/lib/scheme/bfvrns/bfvrns-leveledshe.cpp
+++ b/src/pke/lib/scheme/bfvrns/bfvrns-leveledshe.cpp
@@ -184,7 +184,7 @@ uint32_t FindLevelsToDrop(uint32_t multiplicativeDepth, std::shared_ptr<CryptoPa
     // adding the cushon to the error (see Appendix D of https://eprint.iacr.org/2021/204.pdf for details)
     // adjusted empirical parameter to 16 from 4 for threshold scenarios to work correctly, this might need to
     // be further refined
-    int32_t levels = std::floor((loge - 2 * multiplicativeDepth - 16 - logExtra) / dcrtBits);
+    int32_t levels = std::floor((loge - 4 * multiplicativeDepth - 16 - logExtra) / dcrtBits);
     size_t sizeQ   = cryptoParamsBFVrns->GetElementParams()->GetParams().size();
 
     if (levels < 0)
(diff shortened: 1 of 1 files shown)


yspolyakov (approved):

src/pke/lib/scheme/bfvrns/bfvrns-leveledshe.cpp
```diff
@@ -184,7 +184,7 @@ uint32_t FindLevelsToDrop(uint32_t multiplicativeDepth, std::shared_ptr<CryptoPa
     // adding the cushon to the error (see Appendix D of https://eprint.iacr.org/2021/204.pdf for details)
     // adjusted empirical parameter to 16 from 4 for threshold scenarios to work correctly, this might need to
     // be further refined
-    int32_t levels = std::floor((loge - 2 * multiplicativeDepth - 16 - logExtra) / dcrtBits);
+    int32_t levels = std::floor((loge - 4 * multiplicativeDepth - 16 - logExtra) / dcrtBits);
```
yspolyakov: I believe we should replace 4 with 3 as we get 1 extra bit in the noise estimate for each level, i.e., 2 + 1 = 3.

ZenithalHourlyRate: Changed accordingly. Tested and found out the noise behaves well.

yspolyakov (approved):

yspolyakov:
The fix was revised to 3 * multiplicativeDepth (instead of 4 * multiplicativeDepth). The initial estimate was too conservative.

PR #1097, opened 2025-12-02
Fix evaluation of linear functions with `EvalPoly`

### LINKED ISSUE #1096: Cannot evaluate linear functions using `EvalPoly`
_Originally reported in [the forum](https://openfhe.discourse.group/t/cannot-evaluate-linear-functions-using-evalpoly/2227)._

Commit e21a859ac9e5e05935612e0aba392d61dcf1f62a inadvertently introduced an off-by-one error in `internalEvalPolyLinearWithPrecomp`, which throws an exception when the polynomial to evaluate has degree 1 (it should only throw when the polynomial is of degree 0).

hsanzg:
The `internalEvalPolyLinearWithPrecomp` method contained an off-by-one error that caused it to throw an exception when the polynomial to evaluate was a linear function. This patch fixes that mistake, and adds a test that evaluates a linear polynomial on a CKKS ciphertext.

Fixes #1096


pascoec:
@hsanzg, please apply your changes to the dev branch and resubmit.

hsanzg:
Sorry for the delay, I see you've already created a new branch and merge request targeting `dev`. Thanks!

Issue #110, opened 2022-07-15
Public key generation is slower when the HYBRID key switching method is used


yspolyakov:
`GetParamsPK()` uses extended basis for the HYBRID mode. This is needed to support the PRE functionality. Currently this extended basis mode is used for all HYBRID scenarios, not just those involving PRE.

Once the PRE changes by @sararv22 are merged, the PRE mode data member can be used to restrict the use of the extended basis to HYBRID + PRE scenarios only. 

The use of extended basis increases the `KeyGen` runtime and slightly increases the public key encryption for all cases when the HYBRID key switching method is used.

The changes for the implementation of `GetParamsPK()` need to be made `rlw-cryptoparameters.h`.

@kimandrik  @sarojaduality

PR #1109, opened 2026-01-22
Fix incomplete Issue #110 - change PREMode default to NOT_SET

### LINKED ISSUE #1108: Incomplete Fix for Issue #110 - PREMode default causes public keys to always use Q×P basis when using HYBRID key switching
## Summary
The fix in commit `64ed308` for Issue #110 never worked. The check `m_PREMode != NOT_SET` always passes because `PREMode` defaults to `INDCPA`, not `NOT_SET`.

## The Bug
In `rns-cryptoparameters.h`:
```cpp
if ((m_ksTechnique == HYBRID) && (m_PREMode != NOT_SET))
    return m_paramsQP;
```
In `gen-cryptocontext-params-defaults.h`:
```cpp
constexpr ProxyReEncryptionMode PREMode = INDCPA;  // all three schemes
```
Since `INDCPA != NOT_SET` is always true, public keys always use the extended Q×P basis with HYBRID key switching, whether PRE is needed or not.

## Timeline
- **Aug 3, 2022** (`8e1a341`): `PREMode = INDCPA` added as default
- **Sep 6, 2022** (`64ed308`): Fix attempted with `m_PREMode != NOT_SET` check
- **Present**: Bug persists

## Impact
- Larger public keys when using HYBRID key switching
- No security impact

## Fix
https://github.com/BurakKTopal/openfhe-development/tree/BurakKTopal_issue110

- `gen-cryptocontext-params-defaults.h`: Changed to `PREMode = NOT_SET`
- `pre-buffer.cpp`: Added explicit `parameters.SetPREMode(INDCPA)`

All tests pass, PRE examples work.

## Related
- Original issue: #110
- Attempted fix: `64ed308`

## Note
We do need to look into the effect of this fix on existing, OpenFHE-based PRE implementations.

BurakKTopal:
Fixes issue #110. See complete bug report in issue #1108 

The original fix in commit 64ed308 never worked because PREMode defaulted to INDCPA, not NOT_SET. This caused GetParamsPK() to always return Q×P basis for HYBRID, even when PRE is not needed.

Changes:
- Set default PREMode to NOT_SET in gen-cryptocontext-params-defaults.h
- Update pre-buffer.cpp to explicitly set PREMode=INDCPA


yspolyakov:
@BurakKTopal I agree with the changes, but please create a branch from `dev` in the openfhe-development repository and resubmit the PR. In this case, we can run CI against the branch.

dsuponitskiy:
Replaced with [a new PR](https://github.com/openfheorg/openfhe-development/pull/1122).

PR #1114, opened 2026-02-01
Create an overload for IntMPBootRandomElementGen() taking Ciphertext instead of PublicKey

### LINKED ISSUE #1111: Create an overload for IntMPBootRandomElementGen() taking Ciphertext instead of PublicKey and add it to openfhe-python

dsuponitskiy:


src/pke/lib/scheme/ckksrns/ckksrns-multiparty.cpp
```diff
@@ -170,6 +170,19 @@ Ciphertext<DCRTPoly> MultipartyCKKSRNS::IntMPBootRandomElementGen(std::shared_pt
     return outCtxt;
 }
 
+Ciphertext<DCRTPoly> MultipartyCKKSRNS::IntMPBootRandomElementGen(std::shared_ptr<CryptoParametersCKKSRNS> params,
+                                                                  ConstCiphertext<DCRTPoly>& ciphertext) const {
+    auto ildcrtparams = params->GetElementParams();
```
yspolyakov: It should be created using the element parameters of the first element in the ciphertext, which will be faster than using the cryptocontext element parameters)

yspolyakov (approved):

PR #1164, opened 2026-04-07
isssue1147

### LINKED ISSUE #1147: Some of BFV EvalAdd/Sub* procedures produce incorrect results when ciphertexts have mismatched DCRTPoly formats (EVALUATION vs COEFFICIENT)
See https://openfhe.discourse.group/t/bug-report-bfv-evaladd-produces-incorrect-results-when-ciphertexts-have-mismatched-dcrtpoly-formats-evaluation-vs-coefficient/2304 for more details

pascoec:
Added checks and conversions to
1. EvalAddInPlace
2. EvalSubInPlace
3. EvalMultInPlace for ciphertext and plaintext

### DIFF
--- a/src/pke/lib/schemebase/base-leveledshe.cpp
+++ b/src/pke/lib/schemebase/base-leveledshe.cpp
@@ -89,9 +89,14 @@ Ciphertext<Element> LeveledSHEBase<Element>::EvalAdd(ConstCiphertext<Element>& c
 template <class Element>
 void LeveledSHEBase<Element>::EvalAddInPlace(Ciphertext<Element>& ciphertext, ConstPlaintext& plaintext) const {
     auto& cv = ciphertext->GetElements();
-    auto pt  = plaintext->GetElement<Element>();
-    pt.SetFormat(cv[0].GetFormat());
-    cv[0] += pt;
+    if (cv[0].GetFormat() == plaintext->GetElement<Element>().GetFormat()) {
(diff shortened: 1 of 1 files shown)


src/pke/lib/schemebase/base-leveledshe.cpp
```diff
@@ -172,10 +182,16 @@ Ciphertext<Element> LeveledSHEBase<Element>::EvalMult(ConstCiphertext<Element>&
 
 template <class Element>
 void LeveledSHEBase<Element>::EvalMultInPlace(Ciphertext<Element>& ciphertext, ConstPlaintext& plaintext) const {
-    auto pt = plaintext->GetElement<Element>();
-    pt.SetFormat(Format::EVALUATION);
-    for (auto& c : ciphertext->GetElements())
-        c *= pt;
+    if (plaintext->GetElement<Element>().GetFormat() == Format::EVALUATION) {
```
yspolyakov, line 185: What if the ciphertext is in the COEFFICIENT representation and plaintext in the EVALUATION representation?

pascoec: Then line 187 will throw an error

yspolyakov: It seems like the logic is different from Add/Sub. In my mind, it should check whether both of them are in EVALUATION, adjust if needed, and then do the multiplication

yspolyakov (approved):

PR #1177, opened 2026-04-29
security: branch-free modular primitives on decrypt hot path

BAder82t:
## Summary

This PR removes data-dependent branches from modular arithmetic used during `Decrypt`. Branches whose direction depends on secret values can leak information through timing, so this change replaces them with branchless bitmask-based equivalents.

The new helpers also include a small inline-asm barrier for GCC and Clang to make it harder for the optimizer to rewrite the branchless form back into conditional branches at higher optimization levels.

### New header — `src/core/include/utils/constanttime.h`

Adds four small inline helpers in namespace `lbcrypto::ct`. Each helper implements the same operation as the original conditional logic, but uses a mask derived from the sign bit or borrow bit instead of a branch.

| Helper | Replaces |
|---|---|
| `AddIfNeg(x, m)` | `x < 0 ? x + m : x` |
| `SubIfGE(x, m)` | `x >= m ? x - m : x` |
| `ModSubFast(a, b, m)` | `a - b mod m` |
| `SubIfAboveHalf(x, m, halfQ)` | centered-lift correction when `x > halfQ` |

### Call sites — `src/core/include/math/hal/intnat/ubintnat.h`

The existing public functions are preserved; only their internals change.

- `ModAddFast` / `ModAddFastEq` now call `ct::SubIfGE`
- `ModSubFast` / `ModSubFastEq` now call `ct::ModSubFast`
- `ModMulFastConst` / `ModMulFastConstEq` now call `ct::AddIfNeg`

### NTT butterflies — `src/core/include/math/hal/intnat/transformnat-impl.h`

The NTT butterflies previously had separate GCC and non-GCC implementations, each with its own conditional fixups such as:

```cpp
if (hiVal >= modulus) hiVal -= modulus;
if (loVal < omegaFactor) loVal += modulus;

### DIFF
--- a/src/core/include/math/hal/intnat/transformnat-impl.h
+++ b/src/core/include/math/hal/intnat/transformnat-impl.h
@@ -160,16 +160,8 @@ void NumberTheoreticTransformNat<VecType>::ForwardTransformIterative(const VecTy
 
                 omegaFactor = omega.ModMul(oddVal, modulus, mu);
                 evenVal     = (*result)[indexEven];
-                oddVal      = evenVal;
-                oddVal += omegaFactor;
-                if (oddVal >= modulus) {
-                    oddVal -= modulus;
-                }
-
-                if (evenVal < omegaFactor) {
-                    evenVal += modulus;
-                }
-                evenVal -= omegaFactor;
+                oddVal      = evenVal.ModAddFast(omegaFactor, modulus);
+                evenVal     = evenVal.ModSubFast(omegaFactor, modulus);
 
                 (*result)[indexEven] = oddVal;
                 (*result)[indexOdd]  = evenVal;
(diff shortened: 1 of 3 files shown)


yspolyakov:
We need to benchmark these changes as this codes updates low-level bottleneck operations, e.g., NTT.

yspolyakov:
Hi @BAder82t We did comprehensive benchmarking and unfortunately saw a major slowdown in core polynomial operations (up to 2x). While the idea of achieving constant-time execution in operations that touch sensitive material, such as as private keys and raw data, is certainly reasonable, we think it should be executed differently. For example, one could create a separate version of core subroutines that would be used as part of Decrypt, Keygen, or Encrypt. However, for the server-side execution, we would still use much faster core subroutines, e.g., NTT., as there is no risk for side channel attacks there. So we have created an issue for the constant-time change (for a future release - https://github.com/openfheorg/openfhe-development/issues/1198) and are closing this PR.

PR #1192, opened 2026-05-26
Add API for serializing/deserializing bootstrapping keys

### LINKED ISSUE #353: Add API for serializing bootstrapping keys
We have APIs for SerializeEvalMultKey and SerializeEvalAutomorphismKey. We should add one for bootstrapping keys. Maybe it is possible to serialize bootstrapping keys with the SerializeEvalAutomorphismKey API, but we need to double check.

dsuponitskiy:
1.added API for serializing/deserializing bootstrapping keys
2. fixed a bug in 
static bool DeserializeEvalAutomorphismKey(std::istream& ser, const ST& sertype, const std::string& keyTag,
                                               const std::vector<uint32_t>& indexList)

### DIFF
--- a/src/pke/examples/ckks-bootstrap-keys-serial.cpp
+++ b/src/pke/examples/ckks-bootstrap-keys-serial.cpp
@@ -127,7 +127,8 @@ int main() {
 
     std::ofstream bootstrapKeyOut(DATAFOLDER + bootstrapKeyLocation, std::ios::out | std::ios::binary);
     ErrorCheck(bootstrapKeyOut.is_open(), "Error opening bootstrap eval-key output file");
-    ErrorCheck(serverCC->SerializeEvalBootstrapKey(bootstrapKeyOut, SerType::BINARY, keyPair.secretKey, numSlots),
+    ErrorCheck(serverCC->SerializeEvalBootstrapKey(bootstrapKeyOut, SerType::BINARY, serverCC,
+                                                   keyPair.secretKey->GetKeyTag(), numSlots),
                "Error serializing bootstrap eval keys");
     bootstrapKeyOut.close();
 
@@ -148,6 +149,8 @@ int main() {
                "Error deserializing secret key");
     ErrorCheck(Serial::DeserializeFromFile(DATAFOLDER + ciphertextLocation, clientCiphertext, SerType::BINARY),
                "Error deserializing ciphertext");
+    clientCC->EvalBootstrapSetup(levelBudget, {0, 0}, numSlots);
+    const auto bootstrapKeyIndices = clientCC->GetScheme()->EvalBootstrapKeyMapIndices(clientCC, numSlots);
 
     std::ifstream multKeyIn(DATAFOLDER + multKeyLocation, std::ios::in | std::ios::binary);
     ErrorCheck(multKeyIn.is_open(), "Error opening eval-mult key input file");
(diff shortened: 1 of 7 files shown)


src/pke/examples/ckks-bootstrap-keys-serial.cpp
```diff
@@ -0,0 +1,177 @@
...
+               "Error serializing ciphertext");
+
+    std::ofstream multKeyOut(DATAFOLDER + multKeyLocation, std::ios::out | std::ios::binary);
+    ErrorCheck(multKeyOut.is_open(), "Error opening eval-mult key output file");
+    ErrorCheck(serverCC->SerializeEvalMultKey(multKeyOut, SerType::BINARY), "Error serializing eval-mult keys");
+    multKeyOut.close();
+
+    std::ofstream bootstrapKeyOut(DATAFOLDER + bootstrapKeyLocation, std::ios::out | std::ios::binary);
+    ErrorCheck(bootstrapKeyOut.is_open(), "Error opening bootstrap eval-key output file");
+    ErrorCheck(serverCC->SerializeEvalBootstrapKey(bootstrapKeyOut, SerType::BINARY, keyPair.secretKey, numSlots),
```
yspolyakov: Why is secretKey supplied as an argument? Whoever has access to the bootstrapping key does not need to have access to the secret key.

src/pke/examples/ckks-bootstrap-keys-serial.cpp
```diff
@@ -0,0 +1,177 @@
...
+               "Error deserializing ciphertext");
+
+    std::ifstream multKeyIn(DATAFOLDER + multKeyLocation, std::ios::in | std::ios::binary);
+    ErrorCheck(multKeyIn.is_open(), "Error opening eval-mult key input file");
+    ErrorCheck(clientCC->DeserializeEvalMultKey(multKeyIn, SerType::BINARY), "Error deserializing eval-mult keys");
+    multKeyIn.close();
+
+    std::ifstream bootstrapKeyIn(DATAFOLDER + bootstrapKeyLocation, std::ios::in | std::ios::binary);
+    ErrorCheck(bootstrapKeyIn.is_open(), "Error opening bootstrap eval-key input file");
+    ErrorCheck(clientCC->DeserializeEvalAutomorphismKey(bootstrapKeyIn, SerType::BINARY),
```
yspolyakov: I suggest showing both options of deserializing the bootstrapping keys. The call using DeserializedEvalBootstrapKey will be used more frequently.

src/pke/include/scheme/ckksrns/ckksrns-fhe.h
```diff
@@ -150,6 +150,8 @@ class FHECKKSRNS : public FHERNS {
     std::shared_ptr<std::map<uint32_t, EvalKey<DCRTPoly>>> EvalBootstrapKeyGen(const PrivateKey<DCRTPoly> privateKey,
                                                                                uint32_t slots) override;
 
+    std::vector<uint32_t> EvalBootstrapKeyMapIndices(const PrivateKey<DCRTPoly> privateKey, uint32_t slots) override;
```
yspolyakov: Why is private key supplied as an argument? This function may be called by a party that does not have access to the private key.

src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp
```diff
@@ -299,6 +299,31 @@ std::shared_ptr<std::map<uint32_t, EvalKey<DCRTPoly>>> FHECKKSRNS::EvalBootstrap
...
+std::vector<uint32_t> FHECKKSRNS::EvalBootstrapKeyMapIndices(const PrivateKey<DCRTPoly> privateKey, uint32_t slots) {
+    const auto cryptoParams = std::dynamic_pointer_cast<CryptoParametersCKKSRNS>(privateKey->GetCryptoParameters());
+    if (!cryptoParams)
+        OPENFHE_THROW("Invalid crypto parameters: expected CryptoParametersCKKSRNS");
+
+    auto cc = privateKey->GetCryptoContext();
+    auto M  = cc->GetCyclotomicOrder();
+
+    slots                         = (slots == 0) ? M / 4 : slots;
+    const auto bootstrapRotations = FindBootstrapRotationIndices(slots, M);
```
yspolyakov: I suggest having a suffix of Indices in this variable

yspolyakov (approved):

PR #1211, opened 2026-06-30
Make the rounding in IntBootDecrypt constant-time

### LINKED ISSUE #1198: Make core subroutines that touch sensitive key material or data constant-time
The goal is to provide protection against side channel attacks. For server-side operations, we can still use (faster) core subroutines with varying-time execution. An initial version of the implementation is discussed at https://github.com/openfheorg/openfhe-development/pull/1177.

BAder82t:
Part of #1198.

## Problem

`PolynomialRound`, called from `IntBootDecrypt`, rounds `cs = c1*s + c0`, which depends on the secret key. Its 128-bit loop branched on that value for every coefficient:

- `if (x128 > Q) x128 %= Q;` calls `__umodti3`, whose run time depends on its operands.
- `if ((x128 > Q1quart) && (x128 <= Q3quart))` is the rounding decision itself.

The first version of this PR swapped `ModMulFastConst` and `ModAddFastEq` for masked variants but left both branches in place. Compiled in isolation with clang, those two functions were already branch-free, so the loop did not change. This version removes the branches. It is rebased on current `dev`.

## Change

- Each term of `x128` is below `Q`, so `x128 < 2Q` and `% Q` becomes one masked subtraction (`ct::SubIfGE`).
- The rounding decision becomes a mask, and `qHalf & mask` is always added (`ModAddFastEqCT`).
- New `utils/constanttime.h` with `TopBitMask`, `LessMask` and `SubIfGE`. The bit each mask is built from passes through `ValueBarrier`, the approach of BoringSSL's `value_barrier_w`. Without it, clang on x86-64 turned the rounding mask back into a jump. The barrier is a no-op on MSVC and Emscripten.
- `ModMulFastConstCT` and `ModAddFastEqCT` are added next to the originals, which are unchanged. Server-side code (NTT, EvalMult) is untouched.

Out of scope: the BigInteger fallback in `PolynomialRound` (no double-width native integer, e.g. `NATIVE_SIZE=128`; marked in a comment), and the rest of `IntBootDecrypt` (`c1 * s` and the inverse NTT). `ModMulFastConstCT` is not constant-time where `MultDHi` falls back to `MultDPortable` (documented in its comment); those builds take the BigInteger fallback, not the masked loop.

## Verification

Rounding loop of `PolynomialRound` in the compiled object (`-O3`):

| Build | `__umodti3` calls | Conditional branches in the loop |
|---|---|---|
| before, clang 21 arm64 | 1 | several on `x128`, plus the loop counter |
| after, clang 21 arm64 | 0 | loop counter only |
| after, clang 21 x86-64 | 0 | loop counter only |
| after, GCC 15 arm64 | 0 | loop counter only |

Tests:

- New `INTERACTIVE_MP_BOOT_ROUNDING_2PARTY_ONLY`: with `c1 = 0`, `IntBootDecrypt` rounds a chosen `c0` whose coefficients sit at and next to `Q/4` and `3Q/4`, plus `Q/2`, `0`, `Q-1` and random values, checked against a BigInteger reference. Changing `<=` to `<` in the old loop makes it fail while the existing decrypt test still passes.
- New `UTConstantTime`: the helpers against branching definitions for 32-, 64- and 128-bit words, and the constant-time variants against the originals at edge and random inputs.
- `core_tests` 167/167 and `pke_tests` 2034/2034, including the interactive bootstrapping tests.
- The changed files build with `-Wall -Werror` at `NATIVE_SIZE=128`, at `NATIVE_SIZE=32` (where the masked loop runs with a 64-bit `x128`), and with GCC 15.

`PolynomialRound` alone, 60-bit moduli, single thread, Apple Silicon with clang; before and after produce identical output:

| Ring dimension | Before | After |
|---|---|---|
| 2^16 | 11.9–13.3 ns/coeff | 4.1–4.8 ns/coeff |
| 2^12 | 4.8 ns/coeff | 5.2 ns/coeff |

### DIFF
--- a/src/core/include/math/hal/intnat/ubintnat.h
+++ b/src/core/include/math/hal/intnat/ubintnat.h
@@ -41,6 +41,7 @@
 #include "math/hal/integer.h"
 #include "math/nbtheory.h"
 
+#include "utils/constanttime.h"
 #include "utils/debug.h"
 #include "utils/exception.h"
 #include "utils/inttypes.h"
@@ -756,6 +757,11 @@ public:
         return *this;
     }
 
+    NativeIntegerT& ModAddFastEqCT(const NativeIntegerT& b, const NativeIntegerT& modulus) {
+        m_value = ::lbcrypto::ct::SubIfGE(m_value + b.m_value, modulus.m_value);
+        return *this;
+    }
+
     /**
    * Barrett modulus addition operation.
    *
@@ -1466,6 +1472,13 @@ public:
         NativeInt q = MultDHi(m_value, bInv.m_value) + 1;
         auto yprime = static_cast<SignedNativeInt>(m_value * b.m_value - q * modulus.m_value);
         return {yprime >= 0 ? yprime : yprime + modulus.m_value};
+    }
+
+    NativeIntegerT ModMulFastConstCT(const NativeIntegerT& b, const NativeIntegerT& modulus,
+                                     const NativeIntegerT& bInv) const {
+        NativeInt q            = MultDHi(m_value, bInv.m_value) + 1;
+        SignedNativeInt yprime = static_cast<SignedNativeInt>(m_value * b.m_value - q * modulus.m_value);
+        return {static_cast<NativeInt>(::lbcrypto::ct::AddIfNeg(yprime, static_cast<SignedNativeInt>(modulus.m_value)))};
     }
 
     /**
--- a/src/core/include/utils/constanttime.h
+++ b/src/core/include/utils/constanttime.h
@@ -1,0 +1,82 @@
+//==================================================================================
+// BSD 2-Clause License
+//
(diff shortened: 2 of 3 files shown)


BAder82t:
I updated PR. The first version replaced ModMulFastConst and ModAddFastEq with masked versions, but realised that the loop in PolynomialRound still had the % Q and the if that decides the rounding, so it was not constant-time. Clang also already compiled those two functions without branches, so the first version did not change the loop.

This version removes both branches. The reduction is now one masked subtraction and the rounding decision is a mask. I checked the compiled loop with clang (arm64 and x86-64) and GCC 15, and the only branch left is the loop counter.

I added a test for the rounding boundaries, and core_tests and pke_tests pass. It is about 3x faster at ring dimension 2^16 and about 8% slower at 2^12. The details are in the description.
Since this only covers one part of #1198, I changed from Closes to Part of.

PR #1213, opened 2026-06-30
Retain large allocations on the heap by default

### LINKED ISSUE #1212: Add mallopt static initializer to reduce the number of page faults
OpenFHE ops repeatedly allocate and free large (~0.5–30 MB) polynomial buffers. By default glibc returns each freed buffer to the OS (heap-trim via brk, or munmap), so the next op re-faults those pages from the kernel. Single-threaded page-fault/kernel overhead up to 2× observed on allocation-bound ops e.g., EvalAdd. Modify default policy to keep large allocations on the heap and never auto-trim so buffers are reused instead of re-faulted. Runs once at library load at the cost of higher retained RSS without explicit trim.

pascoec:
OpenFHE operations repeatedly allocate and free large (~0.5–30 MB) polynomial buffers. By default glibc returns each freed buffer to the OS (brk trim / munmap) and re-faults the pages on the next operation; single-threaded, this page-fault/kernel overhead can dominate runtime (~2× on allocation-bound ops such as EvalAdd).

This change sets, once at library load (glibc only):

M_MMAP_MAX = 0 — keep large allocations on the heap
M_TRIM_THRESHOLD = -1 — never auto-return the heap to the OS so freed buffers are reused instead of re-faulted.

Trade-off is higher retained RSS. To bound it, a new AllocTrim() helper explicitly returns free memory to the OS and is called at quiescent teardown points (ClearStaticMapsAndVectors, ReleaseAllContexts).

Notes:
glibc-only; a no-op on other platforms.
AllocTrim() is a performance/RSS utility, not secure erasure — use secure_memset() to wipe sensitive data. It is not async-signal-safe.

### DIFF
--- a/docs/static_docs/Best_Performance.md
+++ b/docs/static_docs/Best_Performance.md
@@ -47,8 +47,8 @@ If an alternative parallelization mechanism is used, e.g., pthreads, C++11 threa
 
 # Memory Allocation Policy
 
-OpenFHE allocates and frees many large polynomial buffers (~0.5–30 MB each) throughout normal operation. Under the default allocator policy, freed buffers of this size are returned to the OS as soon as they are released, resulting in many re-faults when new buffers are requested in subsequent operations.
-When operations run back-toback, this allocate / return / re-fault churn adds substantial page-fault and kernel overhead, resulting in reduced performance.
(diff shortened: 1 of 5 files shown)


yspolyakov:
Please update `Best_Performance.md` to discuss the usage of `AllocTrim()`.

yspolyakov (approved):

PR #1223, opened 2026-07-17
Add a pluggable diagnostic output channel

### LINKED ISSUE #1270: Add a pluggable diagnostic output channel
See https://github.com/openfheorg/openfhe-development/pull/1223 for more details

bnaras:
## Summary

OpenFHE writes diagnostics (warnings, timing dumps, debug notes) directly to `std::cerr` / `std::cout` in a handful of `core`, `binfhe`, and `pke` sources. This PR routes those writes through a thin indirection so an embedding application can redirect them, without changing behavior for anyone who doesn't opt in. Such redirection is necessary, for example, for the `openfhe.R` package to be hosted on CRAN.

It's a no-op by default. With the new `OPENFHE_DEFAULT_LOG_SINK` CMake option `ON` (the default), the diagnostic streams *are* `std::cerr` / `std::cout` and the output is byte-for-byte identical to before.

## Motivation

The direct `std::cerr` / `std::cout` writes don't support two things:

1. **Redirecting diagnostics** to a log file, a GUI, or another logging framework.
2. **Keeping `std::cerr` / `std::cout` out of the compiled archive.** R packages on CRAN may not ship compiled code that writes to stdout/stderr (Writing R Extensions §1.1.3.1); output must go through `REprintf` / `Rprintf`. This isn't possible today without patching the library.

## Design

- **`src/core/include/utils/openfhe_log.h`** — declares `lbcrypto::OpenFHEErrStream()` / `OpenFHEOutStream()` (returning `std::ostream&`) plus `SetOpenFHEErrStream()` / `SetOpenFHEOutStream()` for runtime redirection. Two macros, `OPENFHE_LOG_ERR` / `OPENFHE_LOG_OUT`, expand to the accessors. Returning `std::ostream&` means every existing `operator<<` overload and manipulator (`std::endl`, `std::hex`, …) keeps working, so call sites change only the stream token.

- **`src/core/lib/utils/openfhe_log_default.cpp`** — the default sink, and the *only* translation unit in the library that names `std::cerr` / `std::cout`. Compiled only when `OPENFHE_DEFAULT_LOG_SINK` is defined.

- **`OPENFHE_DEFAULT_LOG_SINK` CMake option, default `ON`.** ON: the default sink is compiled, diagnostics go to `std::cerr` / `std::cout`, and `Set*` lets an embedder redirect at runtime. OFF: the default TU compiles to nothing and the embedder supplies its own definitions of the accessors — keeping `std::cerr` / `std::cout` out of the compiled archive.

- The `std::cerr` / `std::cout` diagnostic sites in `core`, `binfhe`, and `pke` are converted to the macros.

## Behavior / compatibility

- **Default build: no change.** `OPENFHE_DEFAULT_LOG_SINK=ON` reproduces the current behavior exactly.
- **No API break.** Only additions; the macros and accessors are new symbols.
- **Runtime redirect** is available to any embedder via `SetOpenFHEErrStream(std::ostream&)` / `SetOpenFHEOutStream(std::ostream&)`.
- **Build-time swap** (`-DOPENFHE_DEFAULT_LOG_SINK=OFF` + embedder-provided accessors) is for embedders that must exclude `std::cerr` / `std::cout` symbols from the archive.

## Testing

Standalone compile/link/symbol check of both configurations:

- **`ON`:** `OPENFHE_LOG_ERR << … << std::endl` writes to stderr, `OPENFHE_LOG_OUT` to stdout — matching `std::cerr` / `std::cout`.
- **`OFF`:** `nm` confirms the default TU is empty (no `cerr` / `cout` symbols) and converted call sites reference `OpenFHEErrStream()` rather than `std::cerr`; the target links when an embedder supplies the accessors.

## Notes

- No functional change under the default configuration; the seam is purely additive.
- If a fuller logging framework (levels, structured logging) is preferred long-term, this seam is compatible with that direction — the accessors can be backed by a richer implementation later — but the intent here is the minimal, behavior-preserving indirection.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,24 @@ option(WITH_COVTEST "Turn on to enable coverage testing (can be used with g++ on
 option(WITH_NOISE_DEBUG "Use only when running lattice estimator; not for production"  OFF )
 option(WITH_REDUCED_NOISE "Enable reduced noise within HKS and BFV HPSPOVERQ modes"    OFF )
 option(USE_MACPORTS "Use MacPorts installed packages"                                  OFF )
+option(OPENFHE_DEFAULT_LOG_SINK "Compile the built-in diagnostic sink (std::cerr/std::cout); turn OFF to supply your own" ON )
+
+# Diagnostic log sink: when ON (default), compile openfhe_log_default.cpp so
+# OpenFHEErrStream()/OpenFHEOutStream() target std::cerr/std::cout, preserving
+# the historical behavior. When OFF, that translation unit is empty and the
+# embedding application must define those accessors itself, letting an embedder
+# keep std::cerr/std::cout out of the compiled OpenFHE archive (e.g. an R package
+# subject to CRAN's no-output-from-compiled-code rule).
+#
+# OFF is intended for static builds (BUILD_STATIC), where the accessors are
+# resolved when the embedding application links. Combining OFF with a shared
+# build leaves those symbols undefined in the library itself: valid on ELF
+# platforms, which resolve them at load time, but rejected at link time by
+# platforms that require every symbol to resolve when the library is created
+# (macOS, Windows).
+if(OPENFHE_DEFAULT_LOG_SINK)
+    add_definitions(-DOPENFHE_DEFAULT_LOG_SINK)
+endif()
 
 # Set required number of bits for native integer in build by setting NATIVE_SIZE to 64 or 128
 if(NOT NATIVE_SIZE)
--- a/src/binfhe/lib/lwe-pke.cpp
+++ b/src/binfhe/lib/lwe-pke.cpp
@@ -34,6 +34,7 @@
 #include "math/discreteuniformgenerator.h"
 #include "math/ternaryuniformgenerator.h"
 #include "utils/parallel.h"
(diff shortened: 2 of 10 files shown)


pascoec:
please rebase with the dev branch

bnaras:
Rebased onto `dev` — the PR is now a single commit against current `dev`.

Two notes on what changed in the process:

- The rebased diff no longer touches `src/core/lib/math/hal/bigintdyn/ubintdyn.cpp`. That file had `std::cout` in `PrintIntegerConstants()` when this branch was cut from `main`; on `dev` it no longer does, so there is nothing left to convert there. The other eight call-site conversions are unchanged.
- I expanded the comment above the `OPENFHE_DEFAULT_LOG_SINK` option to document the shared-build interaction: `OFF` is meant for `BUILD_STATIC`, where the accessors resolve when the embedding application links. `OFF` with a shared build leaves the accessors undefined in the library itself — fine on ELF platforms, which resolve them at load time, but rejected at link time on macOS and Windows. The default (`ON`) is unaffected on every platform and remains byte-for-byte identical to current behavior.

I verified all eight converted translation units compile and link in a single configuration, and checked the option differentially: with `ON` the built archive still references `cerr`/`cout`, with `OFF` it has zero such references.

PR #1248, opened 2026-08-10
Increased the sparse encapsulation key switching auxiliary modulus

### LINKED ISSUE #1041: Reduce key switching noise in the SPARSE ENCAPSULATION mode of CKKS (functional) bootstrapping
Instead of extending to a single-modulus $p$, extend to auxiliary limbs, e.g., $p_1 p_2$, so that the key switching noise would become comparable to modulus switching noise.

yspolyakov:
## Increase the auxiliary modulus for sparse encapsulation key switching from 60 to ~66 bits

Implements #1041.

### Summary

In `SPARSE_ENCAPSULATED` mode, the key switching to/from the sparse secret at the
bootstrapping modulus raise (`KeySwitchGenSparse` / `KeySwitchSparse`) previously used a
single ~60-bit auxiliary prime (the first prime of the hybrid key switching basis `P`). The
key switching noise `round(q0*e/P)` then dominated the modulus switching noise, which was
visible when the scaling factor is very close to the first modulus (e.g., a 59-bit scaling
factor with a 60-bit first modulus, as used for functional bootstrapping with LUTs of input
bit-size 14 — see the note in `examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md`).

**With this change, the key switching noise is reduced to the modulus switching noise.**
The noise added by the sparse key switch decomposes as

```
(c1 * e)/P'  +  delta_0 + delta_1 * s_new
```

where the second part is the balanced rounding term inherent to any modulus switch. The new
auxiliary modulus `P' = p'_0 * p'_1` (two 33-bit primes, ~66 bits total) exceeds the largest
possible `q0` by ~6 bits, which pushes the key-switching term `(c1*e)/P'` below one per
coefficient — under the rounding floor. Measured at `q0 = 2^60`, `N = 2^12`: the noise added
by the key switch drops from **~2^7.9 to ~2^2.9** (max per coefficient, integer domain), and
2^2.9 matches the theoretical modulus switching (rounding) noise for a Hamming-weight-32
secret. Increasing `P'` further would buy nothing: the rounding term is independent of `P'`.

### Changes

- **`ckksrns-cryptoparameters.{h,cpp}`** — when `secretKeyDist == SPARSE_ENCAPSULATED`,
  `PrecomputeCRTTables` generates the auxiliary basis `P'` of two primes of
  `MAX_MODULUS_SIZE/2 + 3` bits each (33 bits for the standard 64-bit build), distinct from
  all moduli in `Q` and in the hybrid key switching basis `P`, along with the element
  parameters for `{q0, p'_0, p'_1}`, `{p'_0, p'_1}`, `{q0}` and the CRT tables
  (`[P']_{q0}`, `[P'^{-1}]_{q0}`, `[(P'/p'_j)^{-1}]_{p'_j}`, `[P'/p'_j]_{q0}`, the
  `alpha*P'` overflow-correction table, Barrett constants, `1/p'_j`). This mirrors how the
  `ExtendCiphertext` (composite scaling modulus raise) precomputations are handled.
- **`ckksrns-fhe.cpp`**
  - `KeySwitchGenSparse` generates the GHS-style switching key over `q0*P'`
    (`b = -a*s_new + P'*s_old + e`; since `[P']_{p'_j} = 0`, the `P'*s_old` term only
    contributes modulo `q0`).
  - `KeySwitchSparse` extends `c1` from `q0` to `q0*P'` via the centered lift
    (`SwitchModulus`), which is the **exact** CRT basis extension since the source basis has
    a single limb. The mod-down from `q0*P'` to `q0` uses the **exact HPS-style CRT basis
    switch** (`DCRTPoly::SwitchCRTBasis` with the floating-point overflow correction), so no
    `alpha*P'` overflow term is dropped and the only noise added is the centered rounding —
    exactly as in modulus switching. Call-site signatures are unchanged.

### Testing

- Full `pke_tests` suite passes (1900/1900), including all `SPARSE_ENCAPSULATED`
  bootstrapping and functional bootstrapping tests.
- Noise verified in the integer domain (difference of raw decryptions before/after the key
  switch, bypassing the double-precision decode which floors at ~2^-45 for 59-bit scaling):
  max added noise per coefficient is ~2^7.9 before this change and ~2^2.9 after, at
  `q0 = 2^60`, `N = 2^12`, scaling factor `2^59`.

### Notes

- Sparse encapsulation switching keys (automorphism indices `M-4`/`M-2`) generated with
  previous versions are not compatible with this change; bootstrapping keys must be
  regenerated.
- Unit test for regression testing is added.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pascoec (approved):
Reviewed with a focus on whether this changes end-to-end accuracy for 14-bit functional bootstrapping:

**Measured: no accuracy regression — and no accuracy improvement either.**

Configuration: arbitrary-LUT FBT, `PInput` = `POutput` = 16384 (14-bit), `scaleTHI`
= 8000, ring dim = numSlots = 2^17 (131072 slots), order 1, `dnum` 3, `lvlb` {4,4},
`FIXEDMANUAL`, `SPARSE_ENCAPSULATED`, using the fixed unit-test input pattern (which
contains the worst-case inputs 0 and p−1). Comparing `main` @ `ed361af2` against this
branch rebased onto `dev`:

| Q | `main` | this branch | wrong slots (of 131072) |
|---|---|---|---|
| 2^59 — the configuration tests 122–124 use | max \|e\| = 2^43.3484 | 2^43.3483 | 0 / 0 |
| 2^58 — stress | max \|e\| = 2^43.1971 | 2^43.1971 | 432 / 468 |

Worst-slot noise is unchanged to four decimal places. That is expected: at this
configuration the dominant error term is the deterministic Chebyshev residual of the
homomorphic `exp` evaluation, amplified ≈2^13.5 at the two inputs adjacent to the LUT's
wrap discontinuity. It sits roughly 35 bits above the key-switching noise this PR
reduces, so the improvement is real but invisible end to end. The Q58 failures above are
present on `main` too and are unrelated to this change; I'll document them separately.

One framing note, not a blocker:

1. The 2^7.9 → 2^2.9 figure is a per-coefficient microbenchmark at q0 = 2^60, N = 2^12.
   Worth stating explicitly in the PR/issue that it does not translate into an
   end-to-end accuracy gain for functional bootstrapping, so this optimization isn't
   later assumed to have addressed FBT accuracy.

PR #1268, opened 2026-08-19
Support per-dimension gadget bases in FHEW bootstrapping

### LINKED ISSUE #1144: Add FHEW parameterization method and optimizations proposed in https://ia.cr/2025/1892
The goal is to add the method proposed in https://ia.cr/2025/1892 that generates parameters for FHEW/TFHE schemes for a desired probability of failure. The named parameter sets should be regenerated using the new method.

Another goal is to add a new API that generates FHEW/TFHE parameters for a desired probability of failure.

The parameters generated using the new method should be validated (and potentially fine-tuned) using the OpenFHE lattice estimator available at https://github.com/openfheorg/openfhe-lattice-estimator.

HONGDUCK:
Below is a brief summary of the main changes.

The existing parameter structure, which previously supported a single `baseG`, has been extended to support multiple gadget bases. A `gadgetBaseMap` and the corresponding internal `m_baseG_map` were added, while the existing `m_Gpower_map` is used to store the precomputed gadget powers for each base.

`GetBaseG()` and `GetDigitsG()` now take an LWE secret-key coefficient index and return the gadget base and the corresponding number of decomposition digits assigned to that coefficient. To support this throughout the bootstrapping path, the LWE secret-key index is propagated through key generation and accumulator updates so that the appropriate gadget base can be selected during `EvalACC`. A new constructor that takes a `gadgetBaseMap` was also added.

The named parameter sets are now initialized with their corresponding gadget base maps. In addition, the parameter sets with the `LPF_` prefix were updated using the `openfhe-lattice-estimator` so that their failure probability is approximately $2^{-128}$.

The same estimator can also be used to search for parameter sets targeting an arbitrary failure probability. The updated `LPF_` parameter sets in this PR were obtained using this parameter search.

### DIFF
--- a/src/binfhe/include/binfhe-constants.h
+++ b/src/binfhe/include/binfhe-constants.h
@@ -87,12 +87,13 @@ enum BINFHE_PARAMSET {
     STD256Q_LMKCDEY,      // STD256Q optimized for LMKCDEY                                   : 2^(-80)
     STD256Q_3_LMKCDEY,    // STD256Q_LMKCDEY for 3 binary inputs                             : 2^(-70)
     STD256Q_4_LMKCDEY,    // STD256Q_LMKCDEY for 4 binary inputs                             : 2^(-50)
-    LPF_STD128,           // STD128 configured with lower probability of failures            : 2^(-135)
-    LPF_STD128Q,          // STD128Q configured with lower probability of failures           : 2^(-130)
-    LPF_STD128_LMKCDEY,   // LPF_STD128 optimized for LMKCDEY                                : 2^(-150)
-    LPF_STD128Q_LMKCDEY,  // LPF_STD128Q optimized for LMKCDEY                               : 2^(-150)
-    SIGNED_MOD_TEST       // special parameter set for confirming the signed modular         : 2^(-45)
+    LPF_STD128,           // STD128 configured with lower probability of failures            : 2^(-128)
+    LPF_STD128Q,          // STD128Q configured with lower probability of failures           : 2^(-128)
+    LPF_STD128_LMKCDEY,   // LPF_STD128 optimized for LMKCDEY                                : 2^(-128)
+    LPF_STD128Q_LMKCDEY,  // LPF_STD128Q optimized for LMKCDEY                               : 2^(-128)
+    SIGNED_MOD_TEST,      // special parameter set for confirming the signed modular         : 2^(-45)
                           // reduction in the accumulator updates works correctly
+    TOY_MULTI_BASE        // no security; multiple gadget bases for testing                  : not evaluated
 };
 // clang-format on
 std::ostream& operator<<(std::ostream& s, BINFHE_PARAMSET f);
--- a/src/binfhe/include/binfhecontext.h
+++ b/src/binfhe/include/binfhecontext.h
@@ -77,6 +77,9 @@ struct BinFHEContextParams {
     SecretKeyDist keyDist;
 
     double stdDev;
+
+    // number of LWE secret-key coefficients assigned to each gadget base, in ascending base order
+    std::map<uint32_t, uint32_t> gadgetBaseMap;
 };
 
 /**
--- a/src/binfhe/include/lwe-cryptoparameters.h
+++ b/src/binfhe/include/lwe-cryptoparameters.h
@@ -35,6 +35,7 @@
 #include "binfhe-constants.h"
 #include "math/discretegaussiangenerator.h"
 #include "math/math-hal.h"
+#include "math/nbtheory.h"
 #include "utils/serializable.h"
 
 #include <string>
@@ -62,9 +63,8 @@ public:
    * @param baseKS the base used for key switching
    * @param keyDist the key distribution
    */
-    explicit LWECryptoParams(uint32_t n, uint32_t N, NativeInteger q, NativeInteger Q,
-                             NativeInteger q_KS, double std, uint32_t baseKS,
-                             SecretKeyDist keyDist = UNIFORM_TERNARY)
+    explicit LWECryptoParams(uint32_t n, uint32_t N, NativeInteger q, NativeInteger Q, NativeInteger q_KS, double std,
+                             uint32_t baseKS, SecretKeyDist keyDist = UNIFORM_TERNARY)
         : m_q(q), m_Q(Q), m_qKS(q_KS), m_n(n), m_N(N), m_baseKS(baseKS), m_keyDist(keyDist) {
         if (m_n == 0)
             OPENFHE_THROW("m_n (lattice parameter) can not be zero");
@@ -97,8 +97,8 @@ public:
     }
 
     LWECryptoParams& operator=(const LWECryptoParams& rhs) {
-        m_q      = rhs.m_q;
-        m_Q      = rhs.m_Q;
+        m_q = rhs.m_q;
+        m_Q = rhs.m_Q;
         // m_qKS    = rhs.m_qKS;
         m_n      = rhs.m_n;
         m_N      = rhs.m_N;
@@ -130,6 +130,10 @@ public:
 
     uint32_t GetBaseKS() const {
         return m_baseKS;
+    }
+
+    uint32_t GetDigitCountKS() const {
+        return GetDigitCount(m_qKS.ConvertToInt(), m_baseKS);
     }
 
     const DiscreteGaussianGeneratorImpl<NativeVector>& GetDgg() const {
--- a/src/binfhe/include/rgsw-acc-cggi.h
+++ b/src/binfhe/include/rgsw-acc-cggi.h
@@ -75,10 +75,11 @@ private:
    * @param params a shared pointer to RingGSW scheme parameters
    * @param skNTT secret key polynomial in the EVALUATION representation
    * @param m a plaintext
+   * @param index LWE secret-key coefficient index
    * @return a shared pointer to the resulting keys
    */
     RingGSWEvalKey KeyGenCGGI(const std::shared_ptr<RingGSWCryptoParams>& params, const NativePoly& skNTT,
-                              LWEPlaintext m) const;
+                              LWEPlaintext m, uint32_t index) const;
 
     /**
    * CGGI Accumulation as described in https://eprint.iacr.org/2020/086
@@ -88,9 +89,10 @@ private:
    * @param ek1, ek2 evaluation keys for Ring GSW
    * @param a a value to add to the accumulator
    * @param acc previous value of the accumulator
+   * @param index LWE secret-key coefficient index
    */
     void AddToAccCGGI(const std::shared_ptr<RingGSWCryptoParams>& params, ConstRingGSWEvalKey& ek1,
-                      ConstRingGSWEvalKey& ek2, NativeInteger a, RLWECiphertext& acc) const;
+                      ConstRingGSWEvalKey& ek2, NativeInteger a, RLWECiphertext& acc, uint32_t index) const;
 };
 
 }  // namespace lbcrypto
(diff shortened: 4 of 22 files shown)


yspolyakov:
## Review summary

The core mechanism looks correct — per-coefficient gadget bases are propagated consistently between key generation and accumulator evaluation in all three blind-rotation methods, and all pre-existing parameter sets remain behaviorally identical (confirmed by green CI across all backends). However, there is **one issue that should block merge as-is (serialization compatibility)**, plus a test-coverage gap and a few latent hazards worth addressing.

## Blocking issue

**Serialization break without a version bump.** `save`/`load` in `rgsw-cryptoparameters.h` now unconditionally read/write `baseGMap`, but `SerializedVersion()` still returns 1 and `load` doesn't gate the new field on the version. Any `RingGSWCryptoParams` serialized by the current release will fail to deserialize with this PR (JSON: missing-NVP exception; binary: stream misalignment corrupting everything after), and the failure won't be the clean "later library version" message because the version check can't distinguish the formats.

**Suggested fix:** bump `SerializedVersion()` to 2 and read `baseGMap` only when `version >= 2`.

Note that even with that fix, `operator==` now compares `m_baseG_map`, so an old blob (empty map) will compare unequal to a freshly generated context of the same named set — worth deciding how that should behave.

## Verified as correct

- **Index propagation is consistent** in all three methods: CGGI keygen and eval both use coefficient index `i`; DM likewise; LMKCDEY keygen uses `i` and eval passes `indexVec[j]`, which holds the original coefficient index through the permutation map, so each key is decomposed at eval time with exactly the base it was generated under. Buffer sizes (`digitsG2`) are derived from the same `GetDigitsG(index)` on both sides, so no size mismatches.
- **LMKCDEY automorphism keys deliberately stay on the default base** at both keygen (`KeyGenAuto`) and eval (single-poly `SignedDigitDecompose` overload) — a consistent pair, matching the paper's separate `B_at` base.
- **`GetBaseG` cumulative lookup is sound**; the map constructor validates power-of-two bases and nonzero counts. I hand-verified all 43 named-set maps sum exactly to their `latticeParam` (e.g., `LPF_STD128`: 547+9=556; `LPF_STD128_LMKCDEY`: 392+164=556; `LPF_STD128Q_LMKCDEY`: 567+34=601).
- **Sign/function evaluation is unaffected**: the `arbFunc` overload of `GenerateBinFHEContext` still uses the unmapped constructor, so `Change_BaseG`, the time-optimization key map in `BTKeyGen`, and the dynamic base switching in `EvalFloor`/`EvalSign`/`EvalDecomp` behave exactly as before.
- **Security parameters (n, q, Q, N, σ) are unchanged** — only the failure probability moves. An independent check against the paper's Eq. (1) noise model shows the new LPF maps increase blind-rotation variance by ~1.06× (`LPF_STD128Q`) to ~3.0× (`LPF_STD128_LMKCDEY`), directionally and roughly quantitatively consistent with relaxing 2^-135 / 2^-150 to the 2^-128 target. The exact values rest on the openfhe-lattice-estimator run — please record the estimator configuration/commit in the PR or docs so the 2^-128 claims are reproducible.

## Should-fix items

1. **No test exercises a genuinely heterogeneous map.** Green CI only covers single-base maps (all the STD sets); the `LPF_*` sets — the only multi-base ones — appear solely in a benchmark file. A per-index keygen/eval mismatch would pass CI today. Please add at least one gate-bootstrapping unit test per method using an `LPF_*` set (or a custom `BinFHEContextParams` with a two-base map), plus a serialization round-trip of a mapped context.
2. **Coverage validation is asymmetric.** The `count != latticeParam` check exists only in the `BinFHEContextParams` overload; the named-set table is unchecked, and a map covering *more* than n coefficients is silently accepted everywhere (excess entries ignored), while under-coverage throws only lazily at keygen. Moving the check where possible, or asserting coverage at keygen start, would fail fast.
3. **Latent `signEval=true` + map combination.** The new constructor accepts it, but `PreCompute`'s sign-eval branch populates the Gpower map only with the sign-eval base list — a mapped base outside that list would throw `"No GPower found"` at keygen. Unreachable via the public API today; an explicit throw in the constructor would keep it that way.
4. **`Change_BaseG` is silently ignored by per-index paths** when the map is non-empty (`GetBaseG(index)` never consults `m_baseG`). Currently harmless because mapped contexts can't reach the dynamic-base code, but if those paths ever converge, base switching would become a no-op and quietly degrade the noise budget of `EvalSign`/`EvalFloor`. A comment or guard in `Change_BaseG` is cheap insurance.

## Minor notes

- `GetBaseG`/`GetDigitsG` signature changes are source-compatible but ABI-breaking for the binfhe shared library — fine for a minor release, but worth mentioning in release notes alongside the serialization version change.
- The per-call float recomputation in `GetDigitsG(index)` and the O(#bases) map scan are negligible against the NTT cost — no action needed.

pascoec (approved):

pascoec:
The capability is sound and we'll add it now, but the four LPF configurations shipped with this PR
don't really seem to exercise it. We'll regenerate a new set of parameters in #1285 before v1.6 release.

### The configurations don't make use of the capability

icelake, clang18, min of 3 reps, this PR vs its parent. Geomean over 13 gate/keygen rows
is **1.008 at one thread, 1.005 at 36**. Exactly one row moves:

| set | coefficients moved | gate 1t | gate 36t | margin spent |
|---|---|---|---|---|
| LPF_STD128_LMKCDEY | 164/556 (29.5%) | **1.054** | **1.051** | ~25 bits |
| LPF_STD128Q_LMKCDEY | 34/601 (5.7%) | 1.015 | 1.002 | below floor |
| LPF_STD128Q | 20/601 (3.3%) | 1.006 | 1.000 | below floor |
| LPF_STD128 | 9/556 (1.6%) | 1.001 | 1.004 | below floor |

The speedup tracks the fraction of coefficients moved to the larger base, which is the
mechanism working as designed. But three of the four maps move only 1.6–5.7%, so they
gain nothing measurable. The noise side agrees: on fixed code, same build both legs, only
the LPF rows differing, 10 keys x 500 gates pooled, only LPF_STD128_LMKCDEY shows a margin
loss above the measurement floor (~25 bits, against an 11-bit floor set by two untouched
control sets). So one set spends real margin and buys real speed; the other three spend
and buy nothing detectable.

Worth adding for #1285: the LWE key switching is **88%** of total noise variance in
STD192, **68%** in STD128, **30%** in STD128_LMKCDEY and **1%** in STD256 (measured by
feeding `KeySwitch` a noiseless dimension-N ciphertext and reading the phase error). For
the STD128/STD192 families `(qKS, baseKS, digitCount)` is the dominant lever, well ahead
of the gadget base.

### Cleanup

I've also made a number of formatting, cleanup and optimization changes on top — the
per-index parameter table, replacing the `std::optional` accessors with plain overloads,
eager validation of the base map, and a few smaller things. Nothing behavioural beyond
what's called out below.

### Open question: `Change_BaseG` — this is what I'd like input on

`Change_BaseG()` updates `m_baseG`, `m_digitsG` and `m_Gpower`, but not the new per-index
table. The accumulator reads through `GetBaseGParams(index)`, which consults those default
members only when the table is empty — so on a context built with a base map,
`Change_BaseG` mutates state nothing reads. It appears to succeed and changes nothing.

The two features can't currently meet: `Change_BaseG` serves the runtime base switching in
`EvalSign`/`EvalDecomp` and the per-base pregeneration in `BTKeyGen`, both of which need
`m_Gpower_map` populated with 2^14/2^18/2^27, and that only happens under
`PreCompute(signEval=true)` — set at exactly one call site, the arbFunc constructor, which
uses the single-base constructor. Nothing states or enforces that, so a user calling
`GenerateBinFHEContext(BinFHEContextParams)` with a `gadgetBaseMap` and then
`EvalFloor`/`EvalSign` walks straight into it.

I made it **throw when a base map is present**, so the incompatibility surfaces at the
point of misuse rather than silently doing nothing. It has no effect on any path that
works today.

The alternative is that `Change_BaseG` should **rebuild the per-index table uniformly with
the new base** — an override of all per-coefficient assignments. That's coherent, but it
needs gadget powers for the new base, and it isn't obvious a global override is meaningful
once someone has deliberately chosen per-coefficient bases.

**Is the conservative option the intended behaviour, or should we implement the override?**
I'll hold off merging until I hear back.

### Two bugs found and fixed along the way

**1. Digit counts were computed in floating point.**
`digitsG = ceil(log(Q)/log(baseG))` and the equivalent for `baseKS` and `baseR` were
evaluated as doubles. That disagrees with exact integer arithmetic on **1498 of 14087**
boundary neighbourhoods across bases 2–300, in both directions — overcounting wastes key
material and work, undercounting leaves a gadget that doesn't span the modulus. Separately,
`ConvertToDouble()` is itself lossy above 2^53, so no floating-point form is sound for the
54–60 bit moduli in use.

Replaced with an exact integer `GetDigitCount()` in `nbtheory.h` (a shift for power-of-two
bases, a division loop otherwise), used by the RingGSW gadget bases, the AP refreshing base,
the LWE key-switching count, and the DCRTPoly trapdoor and gadget vector. **No shipped
parameter set changes**: 3575 `(modulus, base)` pairs produce identical counts.

**2. The signed digit decomposition lost a carry — this one is ours, from #1238.**
`SignedDigitDecompose` biases the centred coefficient by
`H = (baseG/2) * (baseG^digitsG - 1)/(baseG - 1)` and reads `digitsG` masked windows.
`H` is strictly greater than `baseG^digitsG / 2`, so `Q/2 + H` can exceed `baseG^digitsG`,
and masking the top window discards that carry — the decomposition then reconstructs the
coefficient short by `baseG^digitsG mod Q`. `digitsG = ceil(log_baseG(Q))` does not prevent
this, and the only precondition noted in the code is the unrelated word-width one.

It affects six pre-existing shipped `(set, baseG)` pairs — TOY, TOY_MULTI_BASE, STD128_AP
and STD128_LMKCDEY at base 512, STD192Q_LMKCDEY at base 4096, and SIGNED_MOD_TEST at base
128, whose residual reached 65598 against an allowed 64 — plus the LMKCDEY automorphism
path through the second overload.

It's relevant here because **all four LPF maps pick a second base that triggers it** (512
at 27-bit Q, 32 at 25-bit Q), so any noise measurement of those sets taken before the fix
is confounded. On a configuration with every coefficient at an affected base, removing it
was worth **74 bits** of log2(P_fail).

Fixed in both overloads by leaving the top window unmasked so it absorbs the carry. The
biased value stays below `2*baseG^digitsG`, so that digit is bounded by `2*baseG` — no
extra digit, no change to key sizes or work per gate, and the gate suite measures unchanged.
Every shipped pair now reconstructs with a residual of exactly `baseG/2`.

pascoec:
Following up on the `Change_BaseG` question from the review — we've gone with making
it throw when a base map is present.

How we got there: we ran into this while updating our lattice-estimator parameter
tooling, which turned up a separate keygen bug along the way (`BTKeyGen` handed back a
cached bootstrapping key when you called it a second time with a different secret key —
fixed separately). Digging into that made the `Change_BaseG` interaction concrete
rather than hypothetical.

The deciding factor was `BTKeyGen`'s time-optimization path. It loops over every gadget
base calling `Change_BaseG(k)`, generates a key for each, then switches back. If
`Change_BaseG` instead rebuilt the per-index table uniformly with the new base — the
other option we raised — that loop would wipe out the per-dimension assignment on its
first iteration, generate all the cached keys against a uniform gadget structure rather
than the configured map, and then "restore" to a uniform table that isn't the original
map. A single call would quietly corrupt the context, and nothing would report it.

Throwing also keeps the door open for what we suspect is the real answer if these two
features are ever meant to work together: teaching `Change_BaseG` about the map, so it
moves the whole thing to a coarser set of bases instead of collapsing to one global
base. That's what a multi-base config would need for large-precision `EvalSign` /
`EvalFloor`. Turning a throw into that later costs nothing, whereas the rebuild-uniformly
version would bake in semantics we'd have to undo first. Happy to change in a future issue if you have something else in mind.

PR #1272, opened 2026-08-20
Added FIXEDAUTO/FLEXIBLEAUTO/FLEXIBLEAUTOEXT support to CKKS function…

### LINKED ISSUE #1218: Add FBT support for FLEXIBLE* and FIXEAUTO

yspolyakov:
## Summary

Adds support for the **FIXEDAUTO**, **FLEXIBLEAUTO**, and **FLEXIBLEAUTOEXT** scaling techniques to the CKKS functional bootstrapping capability (`EvalFBTSetup`/`EvalFBT`/`EvalMVB*`/`EvalHomDecoding`), which previously fully supported only FIXEDMANUAL. The FLEXIBLE* modes track the exact level-specific scaling factors, which removes the scaling-factor drift of the FIXED* modes: for the same parameters they achieve smaller output noise (equivalently, correctness can be achieved with a smaller CKKS scaling factor).

Closes #1218

## Design

The FBT input is an imported RLWE ciphertext whose true scale is the RLWE modulus (not a CKKS-tracked scaling factor), while in the FLEXIBLE* modes every scalar/plaintext multiplication encodes constants at the precomputed level-specific scaling factor `GetScalingFactorReal(level)`. The implementation keeps the whole pipeline on that precomputed chain **without consuming any extra levels** (the depth budget returned by `GetFBTDepth` is unchanged across all modes):

- **Declaration + matrix folds.** After `ModRaise`, the raised ciphertext is declared to carry the exact scaling factor of its level. The two residual ratios against the nominal power-of-two scaling factor `2^p` are folded into the homomorphic encoding/decoding matrices at setup time:
  - `scaleEnc *= SF(raised level) / 2^p`, so the values entering the trigonometric Hermite interpolation match the nominal convention exactly;
  - `scaleDec *= 2^p / SF(final level)`, so the decoded RLWE output lands at the exact absolute scale `QPrime/POut` (the output has `1 + lvlsAfterBoot` towers).
- **Initial-scaling correction.** For FLEXIBLE*, the `initialScaling` correction is folded into the existing multiplication by `1/(k*N)` after `ModRaise`, so no extra level is consumed (the FIXED* path is unchanged).
- **FLEXIBLEAUTOEXT (RE-CKKS-DE).** The raised basis excludes the extra modulus (as in `EvalBootstrap`), the raised ciphertext sits at level 1, and the input RLWE ciphertext is imported one level deeper so that it consists of a single tower (see the `extOff` logic in the unit tests).
- **`EvalHomDecoding`.** The public `ModReduce`/`LevelReduce` are no-ops outside FIXEDMANUAL, so the final rescaling now calls the internal rescaling explicitly for the AUTO modes. For FLEXIBLE*, `levelToReduce` is handled with a single scalar multiplication that adjusts the value to the scaling factor of the destination level, followed by a direct tower drop (the same logic as `AdjustLevelsAndDepthInPlace`).

The 128-bit build keeps throwing for FLEXIBLE*, matching regular CKKS bootstrapping.

## Tests

- All 48 existing FIXEDMANUAL FBT tests pass unchanged (`TEST_CASE_FBT` gained a `scalTech` field with a default member initializer, so existing rows are untouched).
- New cases 201–206 / 301–306 / 401–406 cover FIXEDAUTO / FLEXIBLEAUTO / FLEXIBLEAUTOEXT for: arbitrary LUTs (binary and 8-bit, orders 1–2, full and sparse packing, SPARSE_TERNARY and SPARSE_ENCAPSULATED), multi-precision sign, consecutive-level computations in the slots domain, and multi-value bootstrapping.
- New `FBT_NOISE` cases 501–502 evaluate the same LUT under FIXEDMANUAL and under FLEXIBLEAUTO(EXT), measure the actual pre-rounding noise in the output RLWE ciphertext (new `MeasureNoiseBits` helper: centered residual modulo `Q/p`), and assert that FLEXIBLE* yields strictly smaller noise.
- Full `pke_tests` suite passes (1917/1917).

## Measured noise (max over slots, in bits)

The size of the FLEXIBLE* advantage depends on which noise source dominates. The FIXED* scaling-factor drift grows roughly linearly with the ring dimension (the distance of the primes from `2^p` grows with `2N`), while the error of the polynomial approximation of the complex exponential is independent of the ring dimension. As a result, **for smaller ring dimensions the FLEXIBLE* benefit can disappear** (all scaling techniques converge to the same approximation-error floor), **but for the larger ring dimensions used in practice it reappears and grows**, since the drift climbs past that floor.

Binary LUT, `Q = 2^33`, SPARSE_TERNARY, full packing, ring `2^5` (drift-dominated even at a small ring, since the binary path uses the cosine evaluation with a smaller approximation error):

| Mode | Noise (bits) |
|---|---|
| FIXEDMANUAL | 21.0 |
| FIXEDAUTO | 21.0 |
| FLEXIBLEAUTO | **14.0** |
| FLEXIBLEAUTOEXT | **14.2** |

8-bit LUT, `Q = 2^47`, SPARSE_ENCAPSULATED, full packing, FIXEDMANUAL vs FLEXIBLEAUTO across ring dimensions:

| Ring dimension | FIXEDMANUAL | FLEXIBLEAUTO | Gap |
|---|---|---|---|
| 2^5 | 25.1 | 24.1 | 1.0 |
| 2^11 | 30.2 | 24.5 | 5.8 |
| 2^13 | 31.3 | 24.8 | 6.5 |
| 2^14 | 33.1 | 25.0 | 8.1 |
| 2^15 | 33.4 | **25.5** | **7.9** |

The same reappearance is seen for SPARSE_TERNARY at ring `2^15` (both configurations show no gap at ring `2^5`, where they sit on the approximation floor of the degree-58 exponential for the wider `[-25, 25]` range):

| Configuration | FIXEDMANUAL | FLEXIBLEAUTO | Gap |
|---|---|---|---|
| 4-bit, `Q = 2^38` | 32.2 | **27.4** | **4.8** |
| 8-bit, `Q = 2^47` | 34.3 | 33.0 | 1.4 |

Notes from the noise analysis:
- The mode-independent floors were experimentally confirmed to be the Chebyshev approximation error of the complex exponential (degree 46 for SPARSE_ENCAPSULATED, degree 58 for SPARSE_TERNARY): substituting a higher-degree approximation removes the floor and restores the FLEXIBLE* gap at small rings. At practical ring dimensions (`2^15`+) this is unnecessary for SPARSE_ENCAPSULATED with up to 8-bit plaintext moduli, because the ring-dependent noise already exceeds the degree-46 floor. The 8-bit SPARSE_TERNARY configuration remains close to its (higher) floor even at `2^15`, which is why its gap is only ~1.4 bits; a higher-degree approximation for the `[-25, 25]` range would widen it.
- FLEXIBLEAUTO and FLEXIBLEAUTOEXT perform equivalently for FBT (the RE-CKKS-DE fresh-encryption advantage does not engage for imported RLWE inputs).
- FIXEDAUTO is consistently ~2–3 bits worse than FIXEDMANUAL in some configurations (lazy-rescale placement effect; pre-existing behavior, correctness unaffected).

### DIFF
--- a/src/pke/examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md
+++ b/src/pke/examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md
@@ -118,5 +118,8 @@ security.
 - The current multiprecision sign evaluation implementation requires that the digit bit size divides the input bit size.
 - Only sparse secrets are currently supported, with the secret key distributions SPARSE_TERNARY (larger probability of
 failure) and SPARSE_ENCAPSULATED (negligible probability of failure).
-- Only the FIXEDMANUAL mode for rescaling is supported.
+- The FIXEDMANUAL, FIXEDAUTO, FLEXIBLEAUTO, and FLEXIBLEAUTOEXT modes for rescaling are supported (for the 64-bit build).
+The FLEXIBLEAUTO and FLEXIBLEAUTOEXT modes track the exact level-specific scaling factors, which removes the
+scaling-factor drift of the FIXED* modes; hence they achieve smaller noise for the same parameters (equivalently,
+correctness can be achieved with a smaller CKKS scaling factor).
 - MULTIPARTY is not supported.
--- a/src/pke/examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md
+++ b/src/pke/examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md
@@ -122,4 +122,6 @@ failure) and SPARSE_ENCAPSULATED (negligible probability of failure).
 The FLEXIBLEAUTO and FLEXIBLEAUTOEXT modes track the exact level-specific scaling factors, which removes the
 scaling-factor drift of the FIXED* modes; hence they achieve smaller noise for the same parameters (equivalently,
 correctness can be achieved with a smaller CKKS scaling factor).
+Composite scaling (COMPOSITESCALINGAUTO and COMPOSITESCALINGMANUAL) is not supported yet, and neither is NORESCALE;
+`EvalFBTSetup` rejects them.
 - MULTIPARTY is not supported.
--- a/src/pke/examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md
+++ b/src/pke/examples/CKKS_FUNCTIONAL_BOOTSTRAPING.md
@@ -124,4 +124,5 @@ scaling-factor drift of the FIXED* modes; hence they achieve smaller noise for t
 correctness can be achieved with a smaller CKKS scaling factor).
 Composite scaling (COMPOSITESCALINGAUTO and COMPOSITESCALINGMANUAL) is not supported yet, and neither is NORESCALE;
 `EvalFBTSetup` rejects them.
+- The 128-bit build (`NATIVE_SIZE == 128`) is not supported yet; `EvalFBTSetup` rejects it.
 - MULTIPARTY is not supported.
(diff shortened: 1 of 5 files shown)


andreea-alexandru (approved):
Looks good: sound changes and valid bug fixes. I also tested the examples with all modes and added one instance of FLEXIBLEAUTO and captured the necessary change for FLEXIBLEAUTOEXT.

pascoec (approved):

PR #1276, opened 2026-08-21
Adds Fourier-Extension Functional Bootstrapping

### LINKED ISSUE #1207: Add Fourier-Extension CKKS functional bootstrapping from 2026/367
Add the functional bootstrapping method for continuous variables from [2026/367](https://eprint.iacr.org/2026/367) (EUROCRYPT'26) to the CKKS implementation in OpenFHE. Use https://github.com/fyhssgss/FEFBS as the starting point.

The plan for initial steps.
1. Create a feature branch in openfhe-development.
2. Add unit tests for the cases that are supported.
3. Add support for UNIFORM_TERNARY + unit tests for it.
4. Add support for FLEXIBLE* + FIXEDAUTO + unit tests for it.

yspolyakov:
# Fourier-Extension Functional Bootstrapping (FEFBT)

Documentation for [PR #1276 — Adds Fourier-Extension Functional Bootstrapping](https://github.com/openfheorg/openfhe-development/pull/1276) (branch `issue1207-ckks-fefbt` → `dev`), covering the two feature commits by **FYHSSGSS** and the follow-up commits by **Yuriy Polyakov**:

| Commit | Date | Summary |
|---|---|---|
| `fff53d67` | 2026-07-14 | Initial feature: API, CKKS implementation, example, unit tests (+1203 lines across 7 files) |
| `8fc52ae2` | 2026-08-20 | Completion pass: LT mode, `FLEXIBLEAUTO*` support, configurable `lEnc`/`lDec`, depth formula, coefficient reorganization |
| `43630dab` | 2026-09-11 | Rewrite on top of the 1.6.0 StC-first bootstrap (tower dropping, `ModRaiseInPlace`, exact scale folding); `COMPOSITESCALING*` support; composite and CS-vs-FA precision tests |
| `5047d2d6` / `51e16b62` | 2026-09-13/14 | Added, then reverted, a K = 768 (`K_UNIFORMEXT`) regime for FEFBT: measurements for [#1310](https://github.com/openfheorg/openfhe-development/issues/1310) showed the larger K only ever compensated for the old approximate basis extension, so K = 512 suffices (net change: none) |

---

## 1. What this feature is

Standard CKKS bootstrapping (`EvalBootstrap`) refreshes a ciphertext by homomorphically evaluating the *identity* function — its only job is to remove the `q·I` overflow left by raising the modulus, via a sine/cosine approximation of modular reduction.

**Functional bootstrapping** evaluates an arbitrary function *f* **during** the refresh, at essentially the cost of one bootstrap plus one polynomial evaluation. OpenFHE already has one flavor (`EvalFBT`, a digit/LUT-style approach for discrete inputs). FEFBT adds a second flavor aimed at **real-valued (continuous) functions**:

> Approximate *f* by a truncated **Fourier series**. Because every Fourier basis function `e^(2πi·j·x)` is 1-periodic, the unknown integer overflow `I` introduced by modulus raising vanishes automatically — `e^(2πi·j·(x+I)) = e^(2πi·j·x)` — so evaluating the series both decrypts homomorphically *and* applies *f* in a single pass.

The "Fourier extension" part of the name refers to how the coefficients are produced: *f* is fitted on a sub-interval of the period (the message domain), with the fit free to do whatever it needs on the rest of the period. This gives much faster-converging series than a plain Fourier expansion of a non-periodic function.

### Relationship to existing bootstrapping code

As of commit `43630dab`, the implementation mirrors `EvalBootstrapStCFirst` **exactly** and can be characterized by a simple equivalence:

> **FEFBT = the StC-first bootstrap with (a) the correction factor fixed at 0, (b) the exact `2^-deg` shrink *kept* instead of restored — it is precisely the half-period embedding `t = μ/2` of the message into the Fourier-series argument — and (c) the sine stage replaced by the complex-exponential Chebyshev evaluation, double-angle squarings, and the Fourier series via `EvalPoly`, with no post-stage integer restores.**

Because the shrink feeds the exponential's argument and a general *f* is not homogeneous, no correction factor can exist for FEFBT (unlike the sine, whose linear output allows shrink-and-restore), and `deg = round(log2(q0/SF))` must be exactly 1 — **`FirstModSize = ScalingModSize + 1` is now enforced at runtime** with a clear error. A direct consequence of the equivalence is that the composite-scaling matrix folding of the StC-first bootstrap transfers verbatim.

All shared machinery is reused: `CKKSBootstrapPrecom`, `EvalCoeffsToSlots`/`EvalSlotsToCoeffs`, the linear-transform (LT) fallback, `EvalBootstrapKeyGen` for keys, `AdjustCiphertext`, `ModRaiseInPlace` (which internally handles the sparse-encapsulation key-switch sandwich and the composite `ExtendCiphertext` raise), and the radix-optimized `EvalPartialSumInPlace`.

---

## 2. Pipeline

```mermaid
flowchart TD
    A["Input ciphertext (slots encoding, at any level at or above expectedLevel)"] --> B["Drop unnecessary towers: exact scaling-factor adjustment down to expectedLevel"]
    B --> C["SlotsToCoeffs (or LT when levelBudget = {1,1})"]
    C --> D["Sparse REAL packing: rotate-add by slots"]
    D --> E["ModReduce + AdjustCiphertext(1.0): exact SF alignment, no correction shrink"]
    E --> F["ModRaiseInPlace: raw raise / composite ExtendCRTBasis / SE key-switch sandwich"]
    F --> G["Normalize by pre/(kN); sparse: EvalPartialSumInPlace"]
    G --> H["CoeffsToSlots + conjugate-add (take real part)"]
    H --> I["Chebyshev evaluation of the complex exponential (table per key distribution)"]
    I --> J["r double-angle squarings → z = exp(2πi·x); q·I overflow vanishes by periodicity"]
    J --> K["EvalPoly: series Σ cⱼ·zʲ with user Fourier coefficients"]
    K --> L["+ conjugate + 2·c₀ → f(x), real, refreshed ciphertext"]
```

Step-by-step, in `FHECKKSRNS::EvalFEFuncBootstrap` (`ckksrns-fhe.cpp`):

1. **Tower dropping.** The input no longer needs to arrive at an exact level: any ciphertext at or above `expectedLevel = L0 − d·(lvlbDec + 2) − ext` is brought down with the exact per-level scaling factors (the `AdjustLevelsAndDepthInPlace` maneuver), which matters for composite scaling where per-level factors deviate more.
2. **StC first.** The message is moved to coefficient encoding while still at low level (`EvalSlotsToCoeffs`, or `EvalLinearTransform` in LT mode). For sparse REAL packing, a rotate-add by `slots` folds the conjugate half (same as `EvalBootstrapStCFirst`).
3. **Scale alignment.** `AdjustCiphertext(ct, 1.0, lvl)` aligns the scaling factor exactly to the raise level — with correction constant 1, since FEFBT admits no correction shrink (§1).
4. **Modulus raise.** `ModRaiseInPlace` performs the raw single-prime raise, the exact `ExtendCRTBasis` raise for composite degrees above 1, and the sparse-encapsulation two-step key switch (keys `2N−4` / `2N−2`) when applicable.
5. **Normalization.** Multiply by `pre/(kN)` where `pre = 2^-deg = 1/2` for degree 1 (folded into the CtS matrix for composite scaling) and `k = K_UNIFORM` for uniform keys (for sparse keys the `1/K` factor is folded into the CtS matrix at setup, §4). For sparse packing, `EvalPartialSumInPlace` replicates the message.
6. **CtS + real part.** `EvalCoeffsToSlots` followed by adding the conjugate.
7. **Complex exponential.** A Chebyshev series (key-distribution-specific table, §4) evaluates the initial exponential; `r` double-angle squarings then produce `z = exp(2πi·x)` exactly on all branches of the raised message — this is where the integer overflow `I` disappears.
8. **Fourier series.** `EvalPoly(ctxtExp, coefficients)` evaluates `Σⱼ cⱼ·zʲ` (with `c₀` zeroed out and re-added in the clear), then the conjugate is added and `2·Re(c₀)` restored:

   **result = 2 · Re( Σⱼ₌₀..d cⱼ · e^(2πi·j·x) )**

   The user's coefficient vector must therefore satisfy `f(x) = 2·Re(Σ cⱼ e^(2πi·j·x))` on the message domain (see §5 for the convention used by the shipped tables).

---

## 3. Public API

Two methods are added at all three layers (`CryptoContextImpl`, `SchemeBase`/`FHEBase` virtuals, `FHECKKSRNS` overrides):

```cpp
// cryptocontext.h
void EvalFEFuncBootstrapSetup(std::vector<uint32_t> levelBudget = {5, 4},
                              std::vector<uint32_t> dim1        = {0, 0},
                              uint32_t slots                    = 0);

Ciphertext<Element> EvalFEFuncBootstrap(ConstCiphertext<Element> ciphertext,
                                        std::vector<std::complex<double>> coefficients) const;
```

- **`EvalFEFuncBootstrapSetup(levelBudget, dim1, slots)`** — precomputes the StC/CtS (or LT) matrices and stores them in the shared `m_bootPrecomMap[slots]` with `BTSlotsEncoding = true`, mirroring `EvalBootstrapSetup`. The composite folding (`pre = SF0·post/q` residual carried at full precision in the matrices, exact `2^-deg` in `scaleEnc`) is identical to the regular StC-first bootstrap. `levelBudget = {enc, dec}` are the CtS/StC level budgets; `{1, 1}` selects the single-linear-transform (LT) mode. `dim1` is the usual baby-step/giant-step dimension hint; `slots = 0` means full packing (`N/2`).
- **`EvalFEFuncBootstrap(ct, coefficients)`** — performs the pipeline of §2 and returns a refreshed ciphertext encrypting `f(message)`. `coefficients` are the one-sided Fourier(-extension) coefficients of *f*.
- **Keys:** no new keygen — call the standard `EvalBootstrapKeyGen(secretKey, slots)` after setup (rotation, conjugation, and sparse-encapsulation keys are shared with regular bootstrapping, including the reduced single-giant-step rotation-key footprint).

A static depth helper is provided in `FHECKKSRNS` (templated for `int64_t` and `complex<double>` coefficient vectors); commit `43630dab` added the `compositeDegree` parameter, which selects the sparse-encapsulated table switch (§4):

```cpp
static uint32_t GetFEFBTDepth(const std::vector<uint32_t>& levelBudget,
                              const std::vector<VectorDataType>& coefficients,
                              SecretKeyDist skd        = SPARSE_TERNARY,
                              uint32_t compositeDegree = 1);
```

which computes

```
depth = levelBudget[0] + levelBudget[1]          // CtS + StC
      + depth(Chebyshev exp series for skd)      // ~6–7 levels
      + r(skd)                                   // double-angle squarings
      + depth(EvalPoly of the Fourier series)    // ~⌈log2(deg)⌉ + O(1)
```

The example adds `+6` levels of headroom on top of this for post-bootstrap computation and encoding margin.

---

## 4. Supported configurations

| Dimension | Supported | Notes |
|---|---|---|
| Key switching | `HYBRID` only | throws otherwise (same as regular CKKS bootstrapping) |
| Scaling technique | `FIXEDMANUAL`, `FIXEDAUTO`, `FLEXIBLEAUTO`, `FLEXIBLEAUTOEXT`, `COMPOSITESCALINGAUTO`, `COMPOSITESCALINGMANUAL` | `COMPOSITESCALING*` added by `43630dab`, using the same folding as the StC-first bootstrap |
| Modulus sizes | `FirstModSize = ScalingModSize + 1` | enforced at runtime: the `2^-deg` shrink is the half-period embedding and cannot be restored, so `deg` must be exactly 1 |
| Secret key distribution | `SPARSE_TERNARY`, `SPARSE_ENCAPSULATED`, `UNIFORM_TERNARY` | selects the exp-approximation table, see below |
| Word size | `NATIVE_SIZE = 64` only | both setup and eval throw on 128-bit builds; unit tests are compiled out for `NATIVEINT == 128` |
| Slots | full (`N/2`) and sparse | sparse path uses rotate-add before the raise + `EvalPartialSumInPlace` after, same as `EvalBootstrapStCFirst` |
| Input level | any level at or above `expectedLevel` | tower dropping with exact scaling-factor adjustment (`43630dab`) |
| Data type | tested with `REAL` (unit tests) and `COMPLEX` (example) | the evaluated result is the real part by construction |

Per-key-distribution constants (class-scope tables in `ckksrns-fhe.h`; naming is `coeff_exp_<K>_double_<degree>` — approximation range `[−K, K]`, then double-angle, Chebyshev degree):

| SecretKeyDist | Range K | Chebyshev table | Double-angle iterations `r` |
|---|---|---|---|
| `SPARSE_TERNARY` | 28 (`K_SPARSE`) | `coeff_exp_28_double_48` | 3 |
| `SPARSE_ENCAPSULATED`, degree 1 | 16 (`K_SPARSE_ENCAPSULATED`) | `coeff_exp_16_double_23` | 4 |
| `SPARSE_ENCAPSULATED`, composite | 28 (`K_SPARSE`) | `coeff_exp_28_double_48` | 3 |
| `UNIFORM_TERNARY` | 512 (`K_UNIFORM`) | `coeff_exp_512_double_23` | 9 |

- The sparse-encapsulated composite fallback mirrors the `K_SPARSE_ALT` switch of regular bootstrapping (the K = 16 bound is not guaranteed under composite scaling; K = 28 is a valid larger bound).
- **`UNIFORM_TERNARY` with composite scaling is currently limited to degree 2 below ring dimension 2^17** (setup throws otherwise), mirroring the regime where regular bootstrapping keeps K = 512. A K = 768 (`K_UNIFORMEXT`) regime was added and then reverted on this branch: measurements for [#1310](https://github.com/openfheorg/openfhe-development/issues/1310) showed that with the exact HPS basis extension on `dev`, the composite mod-raise overflow is statistically identical to FLEXIBLEAUTO's (max |m + q·I|/q₀ of ~355–420 at ring 2^17 for both), so K = 512 suffices and the larger K was only ever compensating for the old approximate `ExtendCiphertext`. This restriction is expected to be lifted together with the removal of the `K_UNIFORMEXT` switches under #1310.

At setup time the encoding/decoding matrices are scaled so the pipeline needs no extra level for range compression: `scaleEnc` folds `1/K` (sparse variants) and, for composite scaling, the exact `2^-deg` shrink plus the near-1 residual `SF0·post/q`; `scaleDec = 1/pre` telescopes the residual away. This is the same structure as `EvalBootstrapSetup`, so the CtS rotations act on the full-magnitude message and their key-switch noise is not amplified.

---

## 5. Fourier coefficients: convention and provided tables

The runtime computes `2·Re(Σⱼ₌₀..d cⱼ zʲ)`, so a coefficient vector for a real function *f* must satisfy:

- `c₀` real, equal to **half** the DC term of the series;
- for `j ≥ 1`, `cⱼ` is the one-sided coefficient (the `c₋ⱼ = conj(cⱼ)` half is supplied by the conjugate-add).

**Domain normalization.** The encrypted message lives in `[−0.5, 0.5]` (one period). A function on `[−B, B]` is handled by fitting `g(t) = f(2Bt)` for `t ∈ [−0.5, 0.5]` — the example builds inputs in `[−0.5, 0.5)` and compares against `f(2·B·t)`.

Coefficient tables follow the naming pattern `coeff_<function>_<B>_double_<degree>`; the function-specific tables live where they are used:

| Table | Function | Domain | Degree | Location |
|---|---|---|---|---|
| `coeff_identity_1_double_25` | `y = x` (baseline) | `[−0.5, 0.5]` | 25 | `ckksrns-fhe.h` (shipped baseline) |
| `coeff_exp_2_double_29` | `exp(x)` | `[−2, 2]` | 29 | example + unit tests |
| `coeff_sigmoid_8_double_34` | `1/(1+e^(−x))` | `[−8, 8]` | 34 | example + unit tests |
| `coeff_gelu_8_double_44` | GELU (tanh form) | `[−8, 8]` | 44 | example + unit tests |

The coefficients are generated with a SciPy-based tooling repo (Fourier-extension least-squares fit); the generator itself is not part of this PR.

---

## 6. Example

`src/pke/examples/FE-functional-bootstrapping-ckks.cpp` demonstrates the full flow at production scale:

- ring dimension `2^16`, `2^15` slots (full packing), `FLEXIBLEAUTO`, 59-bit scaling / 60-bit first modulus, `SPARSE_TERNARY`, `HYBRID` with 3 digits, `levelBudget = {3, 2}`;
- multiplicative depth chosen as `max(GetFEFBTDepth(...))` over the three target functions, `+6`;
- one setup + keygen, then three `EvalFEFuncBootstrap` calls on the same input ciphertext (exp, sigmoid, GELU), each reporting total time, amortized per-slot time, sample values, and mean precision in bits.

At a reduced ring dimension `2^14` (fits in ~15 GB), the example reaches **33–36 bits** of mean precision across the three functions.

---

## 7. Test coverage

`src/pke/unittest/utckksrns/UnitTestFEFBT.cpp` — parameterized GTest suite (`UTCKKSRNS_FEFBT`), ring dimension `2^12`, depth 26, tolerance `eps = 1e-4`, **23 cases**:

| Axis | Covered values |
|---|---|
| Functions | sigmoid (primary), exp, GELU-tanh |
| Key distributions | `SPARSE_TERNARY`, `SPARSE_ENCAPSULATED`, `UNIFORM_TERNARY` |
| Slots | full (`N/2 = 2048`) and sparse (8) |
| Scaling techniques | `FIXEDMANUAL`, `FIXEDAUTO`, `FLEXIBLEAUTO`, `FLEXIBLEAUTOEXT`, `COMPOSITESCALINGAUTO`, `COMPOSITESCALINGMANUAL` |
| Level budgets | `{3, 2}` and LT mode `{1, 1}` |
| Test types | `FEFBT_ACCURACY`, `FEFBT_POST_ROTATION`, `FEFBT_CS_VS_FA` |

New with `43630dab`:

- **Cases 15–19**: composite-scaling accuracy for all three key distributions, both packing modes, and `COMPOSITESCALINGMANUAL`.
- **Cases 20–23** (`FEFBT_CS_VS_FA`): the same bootstrap is run under `FLEXIBLEAUTO` and `COMPOSITESCALINGAUTO` and the CS precision must come within 1.0 bit of the FA precision — the analog of the `FBT_NOISE_VS_FLEXIBLE` rows of `UnitTestFBT` and of `UnitTestCKKSrnsCompositeScalingVsFlexible`, with the tolerance calibrated from the measured spread (10 runs per case):

  | Configuration | Gap (FA − CS) |
  |---|---|
  | Full packing, `SPARSE_TERNARY`, `{3,2}` | 0.010 ± 0.008 bits |
  | 8 slots (LT), `SPARSE_TERNARY`, `{1,1}` | 0.001 ± 0.007 bits |
  | Full packing, `UNIFORM_TERNARY`, `{3,2}` | −0.219 ± 0.165 bits (CS better) |
  | Full packing, `SPARSE_ENCAPSULATED`, `{3,2}` | −0.102 ± 0.017 bits (CS better) |

  The 1.0-bit tolerance sits more than 5 standard deviations above every measured mean.

The whole suite is compiled out on `NATIVEINT == 128` builds. The full `pke` suite passes with these changes (2050+ tests).

---

## 8. What each commit contributed

**`fff53d67` — "Add Fourier-Extension CKKS functional bootstrapping"**
introduced the complete vertical slice: the two API methods through `cryptocontext.h` → `base-scheme.h` → `base-fhe.h` → `FHECKKSRNS`, the ~280-line CKKS implementation, the example, and the first unit tests.

**`8fc52ae2` — "feat: complete majority of FEFBT implementation"**
- **LT mode**: `levelBudget = {1, 1}` builds single linear-transform matrices instead of the collapsed-FFT decomposition.
- **`FLEXIBLEAUTO`/`FLEXIBLEAUTOEXT` support** (first version), configurable `lEnc`/`lDec`, `GetFEFBTDepth` from actual polynomial degrees, `K_SPARSE_ALT → K_SPARSE`, coefficient reorganization.

**`43630dab` — "Align FEFBT with the 1.6.0 StC-first bootstrap and add COMPOSITESCALING* support"**
- Rewrote `EvalFEFuncBootstrapSetup`/`EvalFEFuncBootstrap` to mirror the current `EvalBootstrapSetup`/`EvalBootstrapStCFirst`, based on the equivalence of §1 (correction = 0, `2^-deg` kept as the half-period embedding, exp + Fourier replacing the sine). The `FirstModSize = ScalingModSize + 1` requirement is now enforced with a clear error.
- Brought in the 1.6.0 machinery: tower dropping with exact scaling-factor level adjustment, `AdjustCiphertext` + `ModRaiseInPlace` (sparse-encapsulation auxiliary basis included), the radix-optimized `EvalPartialSumInPlace`, and composite-aware `lEnc`/`lDec` with the exact matrix scale folding. The bespoke `AdjustCiphertextFEFBT` is removed.
- **`COMPOSITESCALING*` support** with the same folding as the regular StC-first bootstrap; sparse-encapsulated composite falls back to the K = 28 table; `GetFEFBTDepth` gained a `compositeDegree` parameter.
- Tests: composite accuracy cases 15–19 and CS-vs-FA precision-gap cases 20–23 with calibrated tolerances (see §7).

**`5047d2d6` + `51e16b62` — K_UNIFORMEXT regime, added then reverted (net change: none)**
A K = 768 complex-exponential table and the `K_UNIFORMEXT`-style selection were added to lift the uniform-composite restriction, then reverted after the measurements for [#1310](https://github.com/openfheorg/openfhe-development/issues/1310): with the exact HPS `ExtendCiphertext` on `dev`, the composite mod-raise overflow matches FLEXIBLEAUTO's exactly, so K = 512 suffices and the entire `K_UNIFORMEXT` concept (including the switches in regular bootstrapping) is slated for removal under #1310. The revert commit records the measurement summary.

---

## 9. Observations for review

Items 1–4, 7, and 8 from the earlier version of this section (misplaced identity table, dead code in `AdjustCiphertextFEFBT`, stale error messages, 128-bit guard inconsistency, stray `<chrono>` include, `EvalFBTSetupInternal` overlap with the issue-1218 FBT work) were **resolved** by the rebase onto `dev` and commit `43630dab`. Still open:

1. **Shared precomputation map**: setup writes to the same `m_bootPrecomMap[slots]` used by `EvalBootstrapSetup`; configuring both regular bootstrapping and FEFBT for the same slot count on one context silently overwrites one with the other.
2. **Pass-by-value coefficient plumbing**: `std::vector<std::complex<double>>` is copied through all four layers on every call (the mutation of `coefficients[0]` justifies one copy at the innermost layer, not four).


yspolyakov:
The below discussion compares the FLEXIBLEAUTO implementation with the one in the issue-1218 branch. This will be addressed after issue-1218 is merged to `dev`.

## FLEXIBLEAUTO handling: comparison with the issue-1218 FBT design

I compared the FLEXIBLE* support here against the approach recently added for `EvalFBT` on issue-1218 (commit `a1204635`). The two share the same core idea — after ModRaise, *declare* the raised ciphertext onto the FLEXIBLE table-SF chain by re-stamping its scaling factor — but this PR implements only half of that design. Issue-1218 is "declaration **+ fold**": the declaration is made exact by folding the scale ratios into the encoding/decoding matrices. FEFBT has the declaration with no fold, which works only because it silently relies on the scaling primes being very close to 2^p.

### Side by side

**Issue-1218 (`EvalFBT`):**
- After raise: `raised->SetScalingFactor(GetScalingFactorReal(raised->GetLevel()))` — the declaration.
- At setup: `scaleEnc *= SF(raisedLevel) / 2^p` and `scaleDec *= 2^p / SF(finalLevel)` — the fold that makes the declaration *exact*, regardless of how far the primes sit from 2^p. No extra level; the depth budget is identical across all scaling modes.
- FLEXIBLEAUTOEXT: raised basis pops the ext modulus and the input is imported one level deeper (`extOff = 1`).
- AUTO modes use internal rescaling (`ModReduceInternalInPlace`), because public `ModReduce` is a no-op outside FIXEDMANUAL.
- No correction factor needed anywhere.

**This PR (`EvalFEFuncBootstrap`):**
- `AdjustCiphertextFEFBT` does the same declaration (re-stamp from the table) — and nothing else.
- No fold: `scaleEnc = pre/k`, `scaleDec = 1/pre` contain no SF-ratio terms.
- AUTOEXT: pops the ext modulus and adds `+1` to `lDec` — equivalent in spirit to `extOff = 1`; this part matches.
- Setup computes `m_correctionFactor` (EvalBootstrap-style), but `EvalFEFuncBootstrap` never uses it.

**Why the bare declaration still passes the tests:** unlike FBT, which imports an *external* RLWE ciphertext at an arbitrary `initialScaling` (ratio far from 1, so the fold is mandatory), FEFBT's raise boundary is internal — the StC output's true scale is the table SF at the bottom level, and for standard single-prime FLEXIBLE chains every table SF is within ~2⁻⁴⁵ of 2^p, so the declared/true mismatch is ~2⁻⁴⁰: invisible next to CKKS noise. The approaches differ in principle, but the numerical gap only opens when the primes stray from 2^p.

### Suggested changes

1. **Fix the correction-factor clobber (real bug).** `EvalFEFuncBootstrapSetup` unconditionally overwrites the shared member `m_correctionFactor` — which FEFBT never reads, but regular `EvalBootstrap` does. It also uses the old formula (`−0.265·(...)+19.1`, clamp 7–13), while `EvalBootstrapSetup` on this branch now uses the newer spreadsheet-fit, `BTSlotsEncoding`-aware formulas (clamp 7–14 / 6–13). Calling `EvalFEFuncBootstrapSetup` after `EvalBootstrapSetup` on the same context silently degrades regular bootstrapping precision. Since the FEFBT design needs no correction factor (same conclusion as issue-1218), please delete the computation entirely.

2. **Adopt the fold.** All FEFBT levels are static at setup time (the StC bottom level and the raised level are fixed by the pipeline), so folding the true-scale/declared-scale ratio into `scaleEnc` is a one-line setup change, exactly as issue-1218 did. For standard FLEXIBLE chains this changes nothing measurable — its value is that it removes the hidden "primes ≈ 2^p" assumption and is the *prerequisite for COMPOSITESCALING support*, where prime deviations are large. (The commented-out code in `AdjustCiphertextFEFBT` is an attempted composite-aware `EvalMult` adjustment that would burn a level; the fold gets the same effect for free.) With the fold in place, the dead code can be removed and `AdjustCiphertextFEFBT` stays a pure declaration.

3. **Align the double-angle loop with the internal-rescale pattern.** The loop does `cc->EvalSquareInPlace` + public `cc->ModReduceInPlace`; the latter is a no-op for all AUTO modes, so correctness currently rides on `EvalSquare`'s lazy auto-rescale. That happens to preserve the table-SF invariant (the tests confirm it), but `EvalBootstrap` and the issue-1218 FBT path both use `algo->ModReduceInternalInPlace` here deliberately — worth matching for consistency and to keep the noise-degree states predictable.

4. **Consider porting the noise-measurement tests.** The suite currently asserts `eps = 1e-4` accuracy only. Issue-1218 added `FBT_NOISE` cases with a `MeasureNoiseBits` helper and asserted that FLEXIBLE* beats FIXEDMANUAL (~6 bits for the binary LUT). The same instrumentation would show whether FLEXIBLE* is actually buying precision in FEFBT or just passing — the Chebyshev-exp approximation floor can mask the difference, exactly as we found on the FBT side.

One thing that does **not** need changing: the AUTOEXT one-level-deeper handling — `lDec + (st == FLEXIBLEAUTOEXT)` plus the ext-modulus pop already mirrors the issue-1218 `extOff` design, and test cases 13–14 cover it.

pascoec (approved):

PR #1278, opened 2026-08-21
Performance optimization of CKKS encoding, serialization, and threshold-FHE operations

### LINKED ISSUE #1277: SHARE-specific optimizations affecting CKKS and threshold methods

pascoec:
This PR is a targeted optimization pass over the operations that dominate
encrypted-analytics workloads built on CKKS with threshold (multiparty) decryption:
plaintext encoding/decoding, ciphertext (de)serialization, partial decryption, and
joint evaluation-key generation. Changes are bit-exact unless noted; the full core
and pke unit-test suites pass.

## Optimizations

- **CKKS special FFT (encode/decode).** The butterfly twiddle factors depend only on
  (stage, position), so they are now precomputed once per cyclotomic order into
  per-stage contiguous tables, removing a modulo and a gather from every butterfly.
  The complex multiply is written on raw doubles — bit-identical to
  `std::complex operator*` for finite inputs — which eliminates a `__muldc3`
  libcall (and its non-finite fixup branches) per butterfly, and the inverse
  transform's epilogue multiplies by the exact power-of-two reciprocal instead of
  dividing per element.

- **DFT precompute-cache thread safety.** The per-cyclotomic-order precompute map was
  read without synchronization while `Initialize()` inserted under `omp critical`.
  Lookups now take a shared lock (map nodes are reference-stable, so the returned
  reference remains valid after release), and insertion uses a unique lock with
  `try_emplace`.

- **CKKS encode scaling pass.** The encoder called `ceil(log2(|x|))` twice per slot
  just to find the maximum-magnitude component. Since `ceil(log2())` is monotonic,
  it now tracks the running max and takes a single `log2` at the end — identical
  result, and the scaling loop becomes vectorizable.

- **`FitToNativeVector` reduction.** The per-coefficient conversion to RNS residues
  performed a hardware divide (and a poorly predicted sign branch) per element. It
  now uses a word-scaled-reciprocal reduction computed once per call — valid for the
  full 64-bit input range, unlike the modulus-scaled Barrett used elsewhere — with a
  branchless centered correction (gated on `HAVE_INT128`; the legacy loop remains
  for other configurations).

- **Binary deserialization of native vectors.** `load` allocated a bounce buffer with
  `malloc`, read into it, copied element-by-element, and freed it. It now reads
  directly into the vector's storage, symmetric with `save`.

- **`SetElement` rvalue overload.** `CiphertextImpl::SetElement` only accepted a
  const reference, so every `SetElement(std::move(...))` call site silently
  deep-copied a full ring element. The move overload fixes all five call sites,
  including both threshold partial-decrypt paths.

- **Threshold partial decryption (`MultipartyDecryptLead/Main`, base and RNS).** The
  noise-scale multiply is skipped when the scale is 1 (always, for CKKS), the result
  is accumulated in place instead of through temporaries, and the RNS
  flooding-mode branch moves the noise element instead of copying it.

- **Multiparty key-generation cleanups.** Braced-initializer-list arguments to
  `SetPublicElements` silently copied both public-key elements despite `std::move`
  (an initializer_list's backing array is const), in both `MultipartyKeyGen`
  overloads and `MultiAddPubKeys`; these now build the vector with moves.
  `MultiMultEvalKey` skips the noise-scale multiply when the scale is 1 and
  accumulates each digit in place. `MultipartyDecryptFusion` drops an unused
  cryptoparameters cast and starts its accumulation with an out-of-place add rather
  than copy-then-add. `MultiAddEvalSumKeys` delegates to its identical
  automorphism-key counterpart.

- **Parallel joint rotation-key generation.** `MultiEvalAutomorphismKeyGen` now uses
  the same OpenMP pattern as the single-party `EvalAutomorphismKeyGen` (map slots
  pre-created so the parallel loop assigns without a critical section, thread count
  from `OpenFHEParallelControls`), with the index list deduplicated (duplicate
  indices would race on one slot) and the missing-key check hoisted out of the
  parallel region, since exceptions may not cross an OpenMP boundary. Generated
  keys are fresh randomness, as before; sampling order differs from the serial
  version.

## Measured highlights

Single-thread pinned A/B against current dev, minimum over repetitions, depth-0
CKKS at 53-bit scale (the threshold-decryption configuration):

- CKKS encode (`MakeCKKSPackedPlaintext`): **1.55–1.58x** (rings 2^13, 2^14);
  decode: 1.14–1.26x.
- Ciphertext binary deserialization: **1.28–1.55x** across depths.
- Threshold partial decryption (RNS Lead/Main): ~1.06–1.09x (dominated by
  noise-flooding sampling and the NTT).
- Joint rotation-key generation, 64 keys at 8 threads: **38.2 ms -> 8.4 ms
  (4.5x)** — previously flat with thread count, now matching the single-party
  path's scaling.

### DIFF
--- a/src/core/include/math/dftransform.h
+++ b/src/core/include/math/dftransform.h
@@ -38,6 +38,7 @@
 
 #include <complex>
 #include <cstdint>
+#include <shared_mutex>
 #include <unordered_map>
 #include <vector>
 
@@ -121,11 +122,20 @@ private:
         std::vector<uint32_t> m_rotGroup;
         // ksi powers
         std::vector<std::complex<double>> m_ksiPows;
+        // per-stage butterfly twiddles for FFTSpecial/FFTSpecialInv, stored contiguously:
+        // the stage with half-length lenh (a power of two <= Nh/2) occupies [lenh - 1, 2*lenh - 1)
+        std::vector<std::complex<double>> m_fwdTw;
+        std::vector<std::complex<double>> m_invTw;
 
         PrecomputedValues(uint32_t m, uint32_t nh);
     };
     // precomputedValues: key - cyclotomic order, data - values precomputed for the given cyclotomic order
     static std::unordered_map<uint32_t, PrecomputedValues> precomputedValues;
+
+    // guards precomputedValues; map nodes are reference-stable, so readers may use the
+    // returned reference after the shared lock is released
+    static std::shared_mutex& PrecompMutex();
+    static const PrecomputedValues& GetPrecomp(uint32_t cyclOrder);
 
     static void BitReverse(std::vector<std::complex<double>>& vals);
 };
--- a/src/core/include/math/hal/intnat/mubintvecnat.h
+++ b/src/core/include/math/hal/intnat/mubintvecnat.h
@@ -654,12 +654,7 @@ public:
         ar(size);
         m_data.resize(size);
         if (size > 0) {
-            auto* data = reinterpret_cast<IntegerType*>(malloc(size * sizeof(IntegerType)));
-            ar(::cereal::binary_data(data, size * sizeof(IntegerType)));
-            for (::cereal::size_type i = 0; i < size; i++) {
-                m_data[i] = data[i];
-            }
-            free(data);
+            ar(::cereal::binary_data(m_data.data(), size * sizeof(IntegerType)));
         }
         ar(m_modulus);
     }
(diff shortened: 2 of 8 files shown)


andreea-alexandru (commented):
Looks reasonable to me. Two comments to double-check:
* The newly enabled OpenMP loop in MultiEvalAutomorphismKeyGen leaves ModInverse inside the parallel region, so an invalid index (≡ 0 mod 2N) now aborts via std::terminate instead of throwing a catchable exception. It's reachable only with pathological input and mirrors existing house style in base-leveledshe.cpp.
* The FitToNativeVector fast path was applied to the encoding copy but not to the byte-identical slow loop in FHECKKSRNS::FitToNativeVector, which sits on the bootstrapping precompute hot path.

yspolyakov (approved):

PR #1295, opened 2026-09-06
Run BinFHE bootstrapping on 32-bit words at native size 64/128

### LINKED ISSUE #1269: BinFHE: use 32-bit native arithmetic at runtime for parameter sets that fit, even in NATIVE_SIZE=64 builds
## Motivation

BinFHE is roughly **2x faster** when the library is built with `NATIVE_SIZE=32`: on an idle
Xeon Platinum 8360Y (single thread pinned, clang-18, `WITH_NATIVEOPT=ON`, min of 4
repetitions), `EvalBinGate` at STD128/GINX measures ~45.6 ms at `NATIVE_SIZE=32` vs
~93.9 ms at `NATIVE_SIZE=64`. The gap has three sources:

- halved memory traffic streaming the bootstrapping key (gate evaluation is largely
  memory-bandwidth-bound),
- twice the SIMD lanes in the NTT and digit-decomposition loops,
- 32-bit modular loops auto-vectorize even at baseline x86-64 (SSE2 provides
  `pmuludq` 32x32->64 and 32-bit shifts), while several 64-bit idioms need AVX-512
  (there is no packed 64-bit arithmetic shift, unsigned compare, or high multiply below
  it). So the 32-bit advantage is even larger in default builds without `-march=native`.

But `NATIVE_SIZE` is a build-wide choice, and the pke schemes effectively require 64-bit
words. Any deployment that needs both pke and BinFHE in one binary is forced to run BinFHE
at 64-bit. The only workaround today is shipping a second, `NATIVE_SIZE=32` binary just
for BinFHE.

**Proposal:** in a `NATIVE_SIZE=64` build, let BinFHE select a 32-bit internal
representation at parameter-generation time whenever the parameter set's moduli fit the
32-bit regime, i.e. `MSB(Q) <= 28` (the `MAX_MODULUS_SIZE` a `NATIVE_SIZE=32` build
enforces, `basicint.h`). This is the same guarantee an NS32 build gives, so no new
headroom analysis is needed.

## Which parameter sets qualify

25 of the 45 shipped sets have `logQ <= 28`, including **every 128-bit set** (STD128*,
classical and quantum, AP/GINX/LMKCDEY variants, and the LPF_* sets) **and the entire
STD256Q family**: TOY, MEDIUM, STD128_AP, STD128, STD128_3/_4, STD128Q, STD128Q_3, all
STD128*_LMKCDEY, STD256Q/_3/_4 (+ LMKCDEY variants), LPF_STD128*, SIGNED_MOD_TEST.

Staying on the 64-bit path: the STD192*/STD192Q* families (34-39 bits), STD256/_3/_4 and
their non-Q LMKCDEY variants (29 bits), and STD128Q_4 (50 bits). The LWE-side moduli are
never the binding constraint (q <= 4096, q_ks <= 2^18).

## Why this is more tractable than it sounds

1. **The native arithmetic stack is already width-agnostic.** `NativeIntegerT<T>` takes
   its double-width type from the `DataTypes<T>` traits (`uint32_t -> uint64_t`), and the
   templated internals (`mubintvecnat`, `transformnat-impl.h`) contain no width
   conditionals. Only the `NativeInteger` alias makes 64-bit "the" width.
2. **Multiple integer/vector/poly widths coexisting in one binary is an established
   pattern:** `lattice.cpp` already instantiates `PolyImpl` for four vector types
   (`MAKE_POLY_TYPE(NativeVector/M2Vector/M4Vector/M6Vector)`). Adding
   `NativeVectorT<NativeIntegerT<uint32_t>>` and its `PolyImpl` is mechanical, and the
   NTT's static twiddle-table caches are per-instantiation, so the widths don't interfere.
3. **BinFHE is a pure `NativePoly`/`NativeVector` consumer** — no `DCRTPoly` or
   `BigInteger` anywhere in `src/binfhe` — so the blast radius is small and self-contained.
4. **The 32-bit math is already a tested configuration:** the binfhe unit tests pass at
   `NATIVE_SIZE=32` today. Only the coexistence machinery is new, not the arithmetic.

## Design sketch

Two options:

**(a) Template the BinFHE implementation classes on the poly type** and have
`BinFHEContext` choose the instantiation at parameter generation. Architecturally clean,
but it churns the public types (`LWECiphertext` etc. are shared_ptrs to concrete impls in
public headers) and the serialization format (dual cereal registrations, or a canonical
wire format with conversion).

**(b) Keep the public API and wire format 64-bit; make the internal representation
32-bit (suggested).** At `BTKeyGen`/deserialization, convert the bootstrapping key and
keyswitching key once into 32-bit caches, and run blind rotation and LWE key switching on
the 32-bit path:

- The BSK must be *stored* 32-bit, not converted per gate — streaming it is the dominant
  memory traffic, and halved key bandwidth is the largest part of the win.
- Widening on `Serialize()` avoids holding both copies resident.
- Input/accumulator conversion at the blind-rotation boundary is one pass, amortized over
  the hundreds of NTT stages in every gate.
- Public `LWECiphertext`, key types, and serialized blobs are unchanged; existing
  serialized keys keep working.

Small enabling refactors either way:

- `MAX_MODULUS_SIZE` is a build-time preprocessor constant; the second width needs a
  per-type counterpart (e.g. a `MaxModulusSize<T>` trait). `LastPrime`/`FirstPrime` are
  already templated on the integer type.
- Explicit instantiation of the 32-bit vector/poly stack in `NATIVE_SIZE=64` builds
  (possibly gated on binfhe being enabled, though the binary-size addition is modest;
  the header-only NTT comes along implicitly).
- A width seam inside binfhe: `RingGSWCryptoParams`, the three accumulators
  (DM/CGGI/LMKCDEY), and the LWE keyswitch internals take the poly/vector type as a
  template parameter internally; `BinFHEContext` picks the instantiation from `MSB(Q)`.

## Validation

Both paths compute identical exact modular arithmetic (all values fit well under 2^28), so
with a seeded PRNG the 32-bit path is **bit-identical** to the 64-bit path — differential
gate tests are exact, not statistical. Beyond that: the existing binfhe suite routed
through the 32-bit path in a 64-bit build, and an A/B against a true `NATIVE_SIZE=32`
build to confirm the runtime path captures the full measured speedup.

## Notes

- Since dispatch is bit-identical, it can be automatic with an opt-out (or opt-in flag on
  `GenerateBinFHEContext` if preferred).
- The same machinery would benefit the functionalities built on binfhe blind rotation
  (functional bootstrap, scheme switching) under the same modulus predicate.

### LINKED ISSUE #1302: BinFHE's switching key stores slices that can never be read, and adds encryptions of zero
## TLDR

The LWE switching key — by far the largest key object BinFHE builds — is a rectangle: `baseKS`
value slices at every one of `digitCount` digit positions. But the **top** digit position can only
ever take `qKS / baseKS^(digitCount-1)` distinct values, so the remaining slices there are
generated, stored, serialized, and never read. On shipped parameters that is **25.0% of the whole
switching key on the eight `qKS = 2^16` sets and 16.7% on the fourteen `qKS = 2^17` sets** —
488 MiB on STD256Q alone.

Separately, the key keeps a slice for digit value **0** at every position, and unlike DM's
accumulator the LWE key switch *uses* it: an encryption of zero is added whenever a digit is
zero, contributing its noise and a full row of memory traffic for nothing. That is another
1.6-3.1% of the key and about the same share of key-switch noise variance, which is the dominant
noise term in most shipped sets.

Both are consequences of the rectangular layout rather than anything algorithmic, and neither
affects the values the key switch computes.

## The unreachable top slice

```cpp
LWESwitchingKey32Impl(uint32_t N, uint32_t baseKS, uint32_t digitCount, uint32_t n)
    : m_keyA(N * baseKS * digitCount * n), m_keyB(N * baseKS * digitCount) { }
```

The evaluation decomposes `a[i]` in base `baseKS`:

```cpp
for (uint32_t j = 0; j < digitCount; ++j) {
    const auto a0 = static_cast<uint32_t>(atmp % baseKS);
    atmp /= baseKS;
    ...
}
```

so at position `j` the digit is `floor(a[i] / baseKS^j) mod baseKS`. Since `a[i] < qKS`, the
highest position is bounded by `qKS / baseKS^(digitCount-1)`, which is smaller than `baseKS`
whenever `baseKS` does not divide `qKS` exactly at that power. Shipped, with the count of
reachable values at the top position:

| sets | qKS | baseKS | digits | top values of baseKS | dead share of the key |
|---|---|---|---|---|---|
| 8 sets (STD128_3, STD128Q_3, STD256Q, STD192_3, STD192Q_3, and their LMKCDEY twins) | 2^16 | 64 | 3 | **16 of 64** | **25.0%** |
| 14 sets (the `_4` variants, STD256_3/4, STD256Q_3/4, and LMKCDEY twins) | 2^17 | 64 | 3 | **32 of 64** | **16.7%** |
| 20 sets (STD128, STD128Q, STD192, STD192Q, STD256, LPF_*, MEDIUM, AP) | 2^15 or 2^18 | 32 or 64 or 128 | 2-3 | full | 0% |

In absolute terms, at 32-bit storage: STD256Q's switching key is about 1864 MiB, of which
**488 MiB is unreachable**; STD192Q_3 is 1424 MiB with 373 MiB unreachable; STD128Q_3 is 482 MiB
with 126 MiB unreachable. The same fractions apply to the serialized form and to key generation
time, since those slices are sampled like any other.

## The zero-digit slices

DM's accumulator generates only nonzero digit values and skips zero ones at evaluation:

```cpp
for (int32_t j = 1; j < baseR; ++j)      // rgsw-acc-dm.cpp: generation starts at 1
if (a0)                                   // rgsw-acc-common.h: evaluation skips zero
```

The LWE key switch does neither. `accumulateRow` adds `K->RowA(i, a0, j)` unconditionally, so
when `a0 == 0` it adds an encryption of zero: `b = noise + svN[i] * (0 * digitsKS[j])`, i.e. pure
noise. Skipping it would remove, on average, `1/baseKS` of the row additions — 3.1% at
`baseKS = 32`, 1.6% at 64 — together with their noise and their memory traffic. Key-switch noise
is 68-88% of the total variance on the sets we have characterized, so this is a small but free
reduction in the dominant term.

**Does skipping leak which digits are zero?** We do not believe so, but it is the right question
to ask, so here is our reasoning. The digits come from `ctQN->GetA()` — the LWE mask, which the
party performing the key switch already holds in the clear, and which is uniform and independent
of both the secret and the message. So "which digits are zero" is a function of public data;
an observer who could time the operation has the ciphertext and can compute it directly. The
omitted slices are themselves encryptions of zero, and publishing fewer of those cannot help an
attacker. DM's accumulator already skips zero digits of the same vector, so this is the shipped
behaviour elsewhere in the library rather than a new posture.

The one consequence worth flagging: it makes the output noise **data-dependent**, since the
variance would then depend on how many digits happened to be zero. That is harmless for the base
scheme — the noise per row is unchanged and less total noise is strictly better — but threshold
and multiparty constructions sometimes rest on flooding arguments that assume an operation's added
noise is independent of the data. Whether that matters is a question for the owners of those
protocols, which is why it is question 2 below rather than something we would just do.

## What a fix looks like

Give the layout per-position extents instead of a rectangle: keep a cumulative offset table and
have `RowA(i, val, pos)` index through it, so the top position allocates only the values it can
reach. Generation loops to that bound; nothing else changes. For the zero slices, either omit
them and branch at evaluation as DM does, or keep the branchless form — the trade is worth
measuring, because the 32-bit key switch is a lazy accumulation whose speed comes partly from
having no branch in the inner loop, and a 97%-predictable branch may still cost more than the
row it skips.

Wire compatibility: the serialized form is produced by `Widen()`, so the narrow layout is free to
change. The 64-bit key has the same rectangular shape and the same dead slices, so ideally both
change together.

## Questions

1. Is the rectangular layout load-bearing anywhere we have not found — an accelerator backend or
   an external tool that indexes the key directly?
2. Would you want the zero-slice change at all? It trades a branchless inner loop for a
   1.6-3.1% saving, and it makes the key switch's added noise data-dependent as described above.
   Either the performance trade or the multiparty-proof question is a sufficient reason to say
   no, and we would rather ask than assume. The unreachable-top-slice half stands on its own
   either way — it changes nothing about what the key switch computes.

pascoec:
# Run BinFHE bootstrapping on 32-bit words, and the work that made it possible

## What this adds

BinFHE's bootstrapping keys and the operations that consume them now run on 32-bit
words whenever the parameters allow it, inside an ordinary 64-bit build. Nothing about
the scheme changes: the same keys, the same gates, the same noise. What changes is the
width of the machine words the blind rotation and key switch actually execute on, and
the amount of memory the keys occupy while they do it.

Almost every predefined parameter set now has a bootstrapping modulus of 25 to 28 bits and a
key-switching modulus between 2^15 and 2^19 (see the re-selected table below; before it, most
sets were 27 or 28 bits and the rest 29 to 50). Both fit comfortably in 32 bits, yet every
coefficient was stored and multiplied as a 64-bit integer. Halving the word width halves
the resident key material, doubles the number of coefficients in each cache line and each
vector register, and lets the NTT inner loops vectorise at every instruction set rather
than only where 64-bit lanes are available.

Measured against the branch base at the previous parameter table (identical parameters on both
sides) across all 40 GINX and LMKCDEY sets, both compilers, both OpenMP runtimes and 1/8/36
threads — 240 cells, zero regressions — **parameter sets whose keys take the 32-bit path run
2.4 to 4.0 times faster on one thread and up to 7.9 times faster at 36 threads**; sets that
keep a 64-bit refresh key still gain 1.5 to 2.7 times
multi-threaded from the accumulator and scheduling work below. Those figures predate the final
region-scheduling commit, which adds a further 6-11% on the wide LMKCDEY sets at high thread
counts. **Resident key material is about 1.9 times smaller**, and gate outputs are
bit-identical to the 64-bit path throughout.

In absolute terms, with the re-selected table below, **a default STD128 boolean gate takes 17.0
ms on a 36-core machine and 29.1 ms on one thread**, where the released library takes 97.1 and
119.3 ms on that same machine; the fastest LMKCDEY set is at 21.6 ms.

## How it is built

The central idea is that the 32-bit form is an *internal representation*, not a new
scheme, a new key type or a new wire format. Everything an application can observe is
unchanged.

**Qualification is per key and automatic.** A refresh key converts when its modulus fits
the 32-bit accumulator's headroom; a switching key converts when its key-switching
modulus fits the same 28-bit bound the 32-bit kernels are exact for and the lazy
accumulator cannot overflow. Neither depends on the
other, so a parameter set can legitimately end up mixed. One predefined set is:
`LPF_STD256Q_4_AP` carries a 51-bit modulus, so its refresh key stays at native width while its
2^18 key-switching modulus still halves the key switch, which is where the bulk lives. Custom
parameters reach the same case, and the native-size tests build one. A key
whose moduli do not fit is generated at the native width exactly as before, so no parameter
set changes behaviour beyond getting faster.

**Keys are generated directly in the narrow form** rather than built wide and converted.
The 64-bit key is never materialised, so peak memory during key generation drops as well
as steady state. The per-index RGSW encryptions exist once, templated on the polynomial
type, so the narrow and native key generators cannot drift apart.

**The wire format is untouched.** Serialization getters widen a fresh 64-bit copy into the
handle they return, so serialized keys are byte-compatible with existing readers and the
narrow form never escapes the process that built it. The returned copy lives only as long
as the caller holds it.

**The accumulator gained a width-parameterised path** alongside the existing one, with the
32-bit inner products accumulating lazily in a 64-bit word and reducing once per output
coefficient instead of once per digit.

### It compares favourably with a genuine 32-bit build

Running narrow inside a 64-bit build is not a compromise against building the library at
`NATIVE_SIZE=32`: it is faster. Measured at eight threads on one machine, back to back, on the
sets as this branch ships them:

| parameter set | gadget digits | `NATIVE_SIZE=32` | this path, in a 64-bit build | stock 64-bit |
|---|---|---|---|---|
| MEDIUM  | 4       | 12.83 ms | **12.38 ms** | 22.85 ms |
| STD128  | 6 and 4 | 18.13 ms | **17.45 ms** | 33.06 ms |
| STD128Q | 8 and 6 | 22.15 ms | **20.92 ms** | 41.43 ms |

The reason is that the build system forces `HAVE_INT128` off whenever `NATIVE_SIZE` is 32, so a
genuine 32-bit build cannot use a 128-bit reduction at all, on hardware that supports one. This
path lives in a 64-bit build and keeps that option, which is what makes the lazy inner product
worth having. Against that, it pays a narrow-and-widen conversion around each gate that a real
32-bit build does not. The margin grows with the gadget digit count, from 3.5 per cent on the
four-digit set to 5.6 per cent on the eight-digit one, which is the shape the lazy reduction
predicts: its gain scales with the number of digits accumulated, while the conversion cost does
not. At one thread the same three sets give 4.1, 8.7 and 11.4 per cent, and the native-width
column is 2.7 to 2.8x rather than 1.9x, so neither conclusion depends on the thread count. Two
sets now carry two gadget bases, so the column gives both digit counts.

The practical consequence is that BinFHE no longer needs a separate 32-bit build to get 32-bit
performance, which matters because such a build changes the word size for the whole library
rather than for the keys that can use it.

## The predefined parameter table is re-selected

The 41 `STD*`/`LPF*` sets are replaced by 108 sets selected by Duality's lattice-estimator
search against this branch's measured gate times and a noise model validated on it: every
(security level, gate arity) cell at failure targets 2^-64 and 2^-128, for GINX, LMKCDEY and,
new, AP, which gets its own `*_AP` family. Suffixed sets are bound to their method, as the
`*_LMKCDEY` sets already were; the unsuffixed sets are tuned for GINX but stay usable with every
method for backward compatibility, so existing `STD128 + AP` code keeps working (the
failure-probability label applies to the tuned method only). The three `LPF_STD256Q_4` cells
have no admissible configuration at N = 2048; they are the one family selected at N = 4096,
and their rows are marked as such because the noise model's measured range stops at 2048.

Every row but one sits at 25 to 28 bits, so **both bootstrapping keys take the 32-bit path on
every predefined set except `LPF_STD256Q_4_AP`**, whose 51-bit modulus keeps its refresh key at
native width while its key-switching modulus still halves. The sets that previously carried 37-
and 50-bit moduli are where the table pays off; the comparison against the released library is
below.

**The failure-probability labels are measured, not predicted.** Every one of the 108 cells was
certified end to end: eight independent keys per cell, 1250 gates per key, sixteen keys where
eight left the label limited by the width of its own uncertainty and thirty-two where sixteen
did, with no wrong answer in 1088 runs. Each label is the failure probability of the
90th-percentile key, rounded toward zero from the unfavourable end of that measurement's
uncertainty, in whole bits — so it is a value the measurement supports even at the bad end.
**Most comments read weaker than the ones they replace and the parameters behind them did not
get worse.** That rounding rule alone costs a median of three bits and up to twelve; judged
under the previous rule against the previous rows, 89 of the 105 comparable picks are stronger
and 11 weaker, a median gain of two bits. Reviewers diffing the enum comments should read that
decomposition rather than the labels alone.

The enum is now generated from one list, `BINFHE_PARAMSET_LIST`, which also produces the
name table, a `convertToBINFHE_PARAMSET` string converter (the counterpart of core's
`convertToSecurityLevel`) and the method-compatibility mask, so a set is added in one place.
**Enumerator values changed** with the reordering: clients that persisted `BINFHE_PARAMSET`
as an integer must store the name instead. The parameter-set benchmark now covers every
predefined set once; 107 of the 108 build and evaluate correctly at `NATIVE_SIZE=32` on both
compilers, and `LPF_STD256Q_4_AP` keeps a guard because a 51-bit modulus cannot be built at
that word size.

**Against the released library.** Measured on the same rig, clang-18 with native optimisation,
the branch with this table against `main` (`ed361af2`) on all 41 predefined sets both name:
gate time 3.3x faster at one thread and 5.0x at eight or more (range 1.7x to 8.7x), zero
decryption failures over 796 runs, key material within eight percent in aggregate but split by
level, larger at 128 bits where the new picks trade switching-key rows for gate time and smaller
at 192 and 256 bits. Thread scaling from one to eight threads improves from 1.27x to 1.94x.

Running the released library's own parameters through this one splits that: at one thread the
library is worth 2.40x at identical parameters and the parameter change 1.39x. The two are not
independent, because the modulus decides whether the 32-bit path can engage. On the twenty sets
whose released modulus already fits 28 bits the library carries 3.44x of a 3.73x total; on the
thirteen that shiped a 34 to 50 bit modulus the library alone gives 1.37x and the parameter change
2.05x, most of which is those sets reaching the narrow path for the first time.

None of it is bought with failure probability. The released table claims labels from 2^-40
to 2^-270 with no consistent target, and for eighteen of thirty-seven names its claim is
weaker than what this table certifies. Where a controlled comparison exists it is small: the
released library ships one parameter row under both `STD128` and `LPF_STD128`, and against
that single baseline the new 2^-64 pick is 4.10x while the new 2^-128 pick is 3.71x, so the
whole cost of the stronger target is 1.10x. Per-set tables for all of this are in the branch
notes, `.claude/notes/newparams-harness/nptable-results.md`.

## Beyond the width change

**The NTT inner loops now auto-vectorise.** They previously did not, on either compiler,
because a 32-bit induction variable combined with a runtime offset made the dependence
analysis non-affine. Correcting the induction types and compacting the per-stage twiddle
factors let both compilers vectorise the butterflies, and the deferred-reduction schedule
below lets the Cooley-Tukey butterfly skip its conditional subtraction entirely when the
modulus has the headroom to absorb the growth.

**The DM and LMKCDEY accumulators no longer serialise their gadget product.** Each external
product ran one parallel region for the forward transforms surrounded by three serial
phases. They are now a single fused region — worth 1.46x on DM and up to 1.78x on LMKCDEY
at eight threads and above — with the digit decomposition spread across the region's
threads on wide gadgets and the 32-bit inner products blocked over coefficient ranges, so
every phase scales with the region.

**The switching and refresh keys store only the rows a bootstrap can reach (#1302).** Both
digit-indexed keys were rectangles of `B` rows at every digit position, but a coefficient
below the modulus reaches only `(M - 1) / B^(digits - 1) + 1` values at the top position, so
the rest of that position was generated, stored, serialized and never read. For the
switching key that is 25 per cent on 44 of the 108 predefined sets and 12 to 29 per cent on a
further 34; for DM's refresh key, which is the larger key on every AP row, it is 30 to 44 per
cent on 24 of the 36 AP sets. Both keys now hold the top position at its reachable
extent, the key switch and the accumulator are unchanged, and key generation shrinks with
the key: STD128 1172 to 895 MiB, STD128_AP 5417 to 3398 MiB (key generation 15.8 to 9.1 s).
The value-0 rows are gone as well: they encrypt zero, so the key switch now skips zero
digits instead of adding them, which removes `1/baseKS` of the switching key, of the
key-switch additions and of its noise variance, worth up to two bits of failure probability
on the small-base rows the key-capped table favours. Switching keys serialized before this
change are not supported.

**The LMKCDEY regions request one team size throughout.** LMKCDEY was the only method whose
bootstrap alternates two parallel-region sizes — the accumulator asked for digitsG2 threads,
the automorphism key switch for half that — and OpenMP runtimes reconfigure their thread
team whenever consecutive regions request different counts. libgomp charges heavily for
that: up to 4.8x against LLVM's runtime on a saturated 8-core machine with bimodal
run-to-run variance, and 1.5 to 2.7x on shipped parameter sets on a 72-core machine even
with most of its cores idle, where asking for 36 OpenMP threads ran slower than asking for
8. The key switch now requests exactly the accumulator's team size, so the sizes are equal
by construction at every width. That removes the penalty and the variance outright — the two
runtimes land within a few per cent of each other on every shipped set and adding threads is
no longer harmful — while keeping the regions at full width, worth a further 6 to 11 per
cent at 36 threads on both runtimes over capping them. The default thread count now achieves
what previously required hand-tuning OMP_NUM_THREADS per parameter set. The per-dimension
gadget map (#1268) had reintroduced the alternation for LMKCDEY, which visits coefficients
in automorphism order, so a two-base map switched the accumulator's width pseudo-randomly at
every step; every region of a context now requests one width, the widest base's, and the
per-thread decomposition scratch grows to the widest base instead of being reallocated on
every change of digit count. Together these are worth 6 to 9 per cent under LLVM's runtime
and 21 to 37 per cent under libgomp on the two-base LMKCDEY sets and are neutral everywhere
else. The mechanism was isolated with a controlled experiment over all four combinations of
worker-per-chunk ratio and size alternation: equal sizes are healthy even with one chunk per
worker, and differing sizes reproduce the penalty even with two — on both runtimes.

---

# Key points for review

**The default changed.** `BTKeyGen` and `BTKeyLoad` now select the 32-bit form where it
qualifies. Keys generated under this default are not bit-identical to keys generated
before it, because the narrow path samples natively rather than narrowing a wide key. Both
are valid keys and the noise is unchanged, but a test that pins key bytes against a fixed
PRNG seed will differ. Generating keys and then serializing them also holds a transient
widened copy, so that particular sequence peaks higher than it used to even though the
steady-state footprint is much smaller.

**A discrete uniform sampler with exact rejection.** The bounded top-chunk sampler is
replaced by the sliver rejection method designed for DPRIVE: draw the minimum number of
raw words, reject any draw at or above the largest contained multiple of the modulus, and
reduce. It is exactly uniform for every modulus and word width, where the previous
implementation filled its top chunk in a way that rejected a third of draws for some
moduli, and it is 19 to 37 per cent faster. PRNG word generation dominates every sampler
in the library, so this lands on more than BinFHE.

**Correctness fixes carried in this branch.** These are not performance work and deserve
separate attention:

- The excess-H gadget decomposition dropped a carry out of its top window, because the
  bias can push the value past the digit count and the top window was masked rather than
  taking the remaining quotient. This affects ten of the fifty shipped
  (parameter set, gadget base) pairs and is live in the released library. Earlier noise
  measurements on the affected sets are invalid.
- `PreCompute` populated the gadget powers only for the three bases that sign evaluation
  switches between, so any context built through the custom-parameter overload with an
  ordinary gadget base left the vector empty and the first key generation read past its
  end. It is a segmentation fault with no exception, it predates the per-dimension gadget
  base work, and the parameter-set overload was never affected.
- A gadget base large enough to represent the modulus in a single digit leaves the
  external product with no rows, because the approximate decomposition discards the first
  digit. That context used to construct successfully and fail much later inside gate
  evaluation; it is now rejected at construction, naming the base.
- `LWECryptoParams` silently dropped its key-switching modulus and noise generator when
  copied, and its assignment operator additionally dropped the secret key distribution, so
  a copied parameter set reported the wrong key-switching digit count.
- The remaining floating-point digit counts are now exact integer arithmetic.
  `std::log` is not correctly rounded and `ConvertToDouble` is lossy above 2^53, so a
  prime just above a power of two could lose a digit. Verified to change no generated
  parameter set: the two forms agree on all 2271 primes the library produces across bit
  sizes 20 to 60 and seven ring dimensions.
- GINX/CGGI silently computed the wrong function with a Gaussian secret key. Its indicator
  encoding maps any coefficient with `|s| >= 2` onto the same pair as `s == 0`, so those
  coefficients drop out of the blind-rotation exponent sum. The result is a clean, low-noise
  encryption under a *different* key — correct about half the time, and invisible to any
  noise-based check. The combination is now rejected, in the one constructor where both
  parameter objects are visible and again in key generation itself. The manual context
  overload also failed to pass `keyDist` to `LWECryptoParams`, so a caller asking for
  Gaussian secrets silently received ternary ones.
- The LWE inner product overflowed for large moduli. `Encrypt` and `Decrypt` accumulated
  across the whole dimension before a single reduction, letting the accumulator reach
  `dim * q` — past a 32-bit word for a ciphertext held at the ring modulus. It never changed
  a decrypted bit, since the corruption is about 10^5 against a decision threshold of
  1.7e7, but it shifted the phase by a multiple of 2^32, silently invalidating any noise
  measurement taken on such a ciphertext.
- `BinFHEScheme::Bootstrap` added its rounding constant using the *input* ciphertext's
  modulus rather than the one it had just switched to.
- `LWECiphertextImpl::SetModulus` reduced the two halves of one ciphertext under different
  interpretations: the vector `ModEq` applies a centered correction when the modulus shrinks
  by less than two, while the integer `ModEq` reduces plainly. No shipped path reaches the
  disagreement — BinFHE's only shrink is exactly two-fold, which takes the plain branch for
  both halves — but the two must not be able to diverge.
- Three memory-safety holes on public entry points are closed. `KeySwitch` indexed the
  switching key by digit value with only a length check, so a ciphertext still at the ring
  modulus read past the compacted top position; it now requires the ciphertext modulus not
  to exceed the key-switching modulus. The DM and LMKCDEY accumulators dereferenced an empty
  refresh-key slot, or indexed the discrete-log table out of range, for a ciphertext whose
  modulus differs from `q`, which `EvalBinGate` lets through when that modulus is neither
  `q` nor `Q`; both schedules now require equality. `BootstrapGateCore` looped forever on a
  ciphertext modulus above `2N`; it now requires the modulus to divide `2N`, the assumption
  its sparse embedding already documents, and the parameter constructor checks the same for
  `q`. The 32-bit switching-key qualification also admitted moduli up to 32 bits although
  the Shoup lane in its generator is exact only to about 2^31; it now uses the same
  `MAX_MODULUS_SIZE32` cap as the refresh key. `LWECryptoParams::operator==` now compares
  the key-switching modulus, the key distribution and the key-switching noise as well.
- Parameter combinations BinFHE cannot represent are now rejected at construction instead of
  failing later, which matters for automated parameter search.

**Testing.** Gate outputs are bit-identical to the 64-bit path across every gate, every
predefined parameter set and all three bootstrapping methods. All 108 predefined sets
evaluate correctly at 1, 8 and 36 threads under both runtimes, and all but `LPF_STD256Q_4_AP`
at `NATIVE_SIZE=32` under both compilers. A new unit test file covers
BinFHE at the build's true native word size, because the rest of the suite exercises
whatever the defaults do and therefore no longer reaches it, and covers the conversions
between the two representations: narrowing an existing key must not change a ciphertext
bit, and the serialization getters must widen without the context giving up its own
narrow keys.


yspolyakov (approved):

yspolyakov:
Below are results of regression automated analysis using Claud Code (Fable v5.1).

## Regressions that should block or be documented before merge

- **Old serialized switching keys are silently misread.** The 64-bit switching key dropped its value-0 row and compacted the top digit, so `KeySwitch` now indexes `refB[a0 - 1]` instead of `refB[a0]` in `src/binfhe/lib/lwe-pke.cpp:363`. `SerializedVersion()` in `src/binfhe/include/lwe-keyswitchkey.h:130` is still 1 and `load()` checks nothing structural. A key serialized before this PR loads without error and produces wrong gate outputs. The 32-bit narrowing constructor does throw on the mismatch, but only when the key qualifies for narrowing. The large-precision context used by CKKS scheme switching has qKS = 2^35, so it takes the silent path. Fix: bump the version and reject version 1, or check the per-index row count in `KeySwitch`.

- **`STD128_AP` and every `*_AP` set no longer accept GINX.** The old compatibility check allowed `STD128_AP` with AP or GINX. The new mask binds it to AP only, so `GenerateBinFHEContext(STD128_AP)` with the default method now throws. The PR's own deep tests were changed to use `STD128_AP, AP`, which hides this.

- **Nearly every pre-existing named parameter set changed.** Only `TOY`, `TOY_MULTI_BASE`, `MEDIUM`, and `SIGNED_MOD_TEST` are byte-identical. The rest changed n, moduli, or gadget bases. Notable: `STD192*` and `STD256*` sets moved q from 2048 to 4096, all `*_LMKCDEY` sets moved q from 1024 to 2048 and autoKeys from 10 to 40, and `STD256*_LMKCDEY` switched from ternary to Gaussian secrets. Ciphertexts and keys produced with the old table are not interoperable with the new one, and documented failure probabilities for several sets dropped (for example `STD128` from 2^-135 to 2^-68). Users who relied on the old LPF-class rate under `STD128` need to move to `LPF_STD128`.

- **`LWECryptoParams::operator==` now compares `m_keyDist`, which is not serialized.** After deserialization the field is always `UNIFORM_TERNARY`, so a Gaussian context compares unequal to its own deserialized copy. The serialization tests only cover `TOY`, so this is not caught.

## API and behaviour changes worth flagging

- **`GetRefreshKey()` and `GetSwitchKey()` return by value.** Binding a non-const reference to them no longer compiles. When the context holds 32-bit keys, each call widens a full 64-bit copy, so repeated calls allocate the entire key every time and pointer identity is gone.
- **`BTKeyLoad` now mutates and frees by default.** It narrows the loaded keys, resets the 64-bit handles, clears the monomial table, and trims the allocator. Previously it was a plain assignment.
- **Key generation is not bit-reproducible against the old library** under a fixed seed, because the 32-bit path samples natively. The PR notes this.
- **Custom-parameter contexts now honour `keyDist` for the LWE key.** Previously Gaussian requests silently produced ternary secrets. GINX with Gaussian now throws at construction. Both are fixes but change results for existing custom setups.
- **New exceptions replace previously wrong results** for `EvalFunc` and `EvalFloor(roundbits>0)` on AP or LMKCDEY contexts, and for a gadget base that leaves fewer than two digits.
- **Shared parameters can lose their monomial table.** `BTKeyGen` releases the 64-bit monomials when all keys are 32-bit, and 64-bit `EvalAcc` throws if they are missing. A copied `BinFHEContext` holding a 64-bit key could hit this after the other copy regenerates keys. Nothing in-tree does this.

## Core changes affecting all schemes

- **Lazy forward NTT.** Values grow by 2q per stage and the fold chain reduces correctly for log2 N up to 31, with the guard `(2·stages+3)·q` fitting the word. One precondition tightened: inputs must be in [0,q), where the old kernel tolerated [0,4q). Every in-tree caller reduces first. Typical 59- and 60-bit CKKS primes do not take the lazy path at N ≥ 2^12, so most CKKS runs use the unchanged kernel.
- **Discrete uniform sampler.** The sliver-rejection method is exactly uniform. It changes the random stream, so seeded outputs differ from the old library. Rejection can approach 50% for moduli just above a 32-bit boundary, which the "a few percent" comment understates, but the old sampler had the same behaviour.
- **Float-to-integer helper rewrites** in packed encoding, CKKS multiparty, and composite prime generation are value-identical for the prime moduli they see.

## Scheme switching is unaffected

The CKKS-to-FHEW path uses the untouched custom context overload. Both bootstrapping keys stay 64-bit there because qKS = 2^35 exceeds the 28-bit limit, so the comment added at `src/pke/lib/scheme/ckksrns/ckksrns-schemeswitching.cpp:1342` claiming the switching key narrows is wrong, though harmless. The new modulus guards are never tripped on that path, and the `uint64_t{1} << logQ` change is a real fix for logQ ≥ 31.

## Testing

Built the branch (Release, OpenMP) and ran the core (167), BinFHE (145), and PKE (2031) unit-test suites: all pass. The core and BinFHE libraries also compile at `NATIVE_SIZE=128`.

PR #1319, opened 2026-09-16
Keep precomputed powers intact in linear series evaluation (#1312)

### LINKED ISSUE #1312: `EvalMVB`/`EvalMVBNoDecoding` corrupt the shared powers for coefficient vectors of degree below 5
## Summary

CKKS functional bootstrapping's multi-value path lets a caller bootstrap once with
`EvalMVBPrecompute` and then evaluate several functions against the returned
`seriesPowers`. When the Hermite trigonometric coefficient vector has degree below 5,
`EvalMVB` and `EvalMVBNoDecoding` scale those powers **in place**. The caller's
`seriesPowers` is left modified, so the first evaluation is correct and every later
evaluation against the same precomputation silently returns wrong values.

This is live rather than theoretical: `p = 4, order = 1` (2-bit digits, first-order
interpolation) is an ordinary configuration that produces degree 3. Single-function
`EvalFBT` is unaffected, since it builds and consumes its powers within one call.

Found while reviewing PR #1276, which is unrelated to this code path.

## Mechanism

`FHECKKSRNS::EvalMVBNoDecodingInternal` (`src/pke/lib/scheme/ckksrns/ckksrns-fhe.cpp:4118`)
reaches `EvalPolyWithPrecomp` at three sites — `:4182` and `:4184` for the real and
imaginary channels of the fully packed case, `:4225` for the sparse case.

`AdvancedSHECKKSRNS::EvalPolyWithPrecomp`
(`src/pke/lib/scheme/ckksrns/ckksrns-advancedshe.cpp:492`) dispatches on degree. At
degree 5 and above it uses the Paterson-Stockmeyer path, which takes the powers by const
reference and is safe to share. Below 5 it uses `internalEvalPolyLinearWithPrecomp`,
which scales the powers as it accumulates:

```
src/pke/lib/scheme/ckksrns/ckksrns-advancedshe.cpp:344

            cc->EvalMultInPlace(powers[i - 1], coefficients[i]);
```

`LeveledSHECKKSRNS::EvalMultCoreInPlace` mutates `ciphertext->GetElements()`, i.e. the
pointee rather than the handle.

The `std::make_shared<seriesPowers<DCRTPoly>>(ciphertexts->powersRe, ...)` repackaging at
`:4168` and `:4218` does not guard against this. It copies the
`std::vector<Ciphertext<DCRTPoly>>`, and the copied `shared_ptr`s address the same
`CiphertextImpl` objects, so the mutation reaches the caller. That repackaging exists for
an unrelated reason: `EvalPolyWithPrecomp` reads only `powersRe`, so the imaginary
channel has to be moved into that field in order to be evaluated.

The `digitBitSize == 1 && order == 1` branch already guards the same hazard correctly,
with a deep copy at `:4147`, `:4148` and `:4202`, which suggests the general path is an
oversight rather than a deliberate trade.

## Reachability

Degrees returned by `GetHermiteTrigCoefficients`
(`src/core/include/math/hermite.h:58`), measured directly:

| p  | order | degree | path                     |
|----|-------|--------|--------------------------|
| 2  | 1     | 1      | special-cased, safe      |
| 2  | 2     | 3      | **affected**             |
| 2  | 3     | 3      | **affected**             |
| 4  | 1     | 3      | **affected**             |
| 4  | 2     | 6      | safe                     |
| 4  | 3     | 7      | safe                     |
| 8  | 1     | 7      | safe                     |
| 8  | 2     | 12     | safe                     |
| 16 | 1     | 15     | safe                     |

Everything at `p >= 8` is unaffected, which is the likely reason this has gone unnoticed:
the shipped example exercises the larger plaintext moduli.

## Impact

No exception and no accuracy warning. The first `EvalMVB` or `EvalMVBNoDecoding` call
against a given precomputation returns correct values; the second and later calls return
wrong ones. Multi-value bootstrapping - evaluating several functions on one bootstrapped
ciphertext - is precisely the use case that breaks.

## Proposed fix

Hand `internalEvalPolyLinearWithPrecomp` deep copies when the powers are meant to be
reused, mirroring the existing degree-1 guard:

```cpp
std::vector<Ciphertext<DCRTPoly>> copies;
copies.reserve(k);
for (uint32_t i = 0; i < k; ++i)
    copies.emplace_back(ciphertexts->powersRe[i]->Clone());
```

Alternatively, make `internalEvalPolyLinearWithPrecomp` non-mutating at the source. That
fixes every caller at the cost of one ciphertext copy per term, on the path that is
already the cheaper of the two. PR #1276 takes the first approach in
`EvalFEFuncBootstrapWithPrecomp`, for this same reason.

## Proposed regression test

At `p = 4, order = 1`: precompute once, evaluate function A, evaluate function B, then
re-evaluate function A and assert it still matches its expected values. The existing
multi-value tests would not catch this at their current plaintext moduli.

## Notes

The degree table above was produced by calling `GetHermiteTrigCoefficients` directly, not
by reading the formula. The rest of the analysis is from the call path; an end-to-end
failing FBT reproducer has not been built.

pascoec:
Below degree 5, EvalPolyWithPrecomp and EvalChebyshevSeriesWithPrecomp evaluated the series straight from the power (or Chebyshev) basis and scaled the basis ciphertexts in place. Any caller sharing a precomputation across several series got a correct first result and silently wrong ones afterwards. EvalMVB/EvalMVBNoDecoding hit this for Hermite coefficient vectors of degree 3 (p = 4 at order 1, p = 2 at orders 2 and 3); the seriesPowers repackaging in EvalMVBNoDecodingInternal shares the same CiphertextImpl objects, so the corruption reached the caller's powers.

Scale each lower-order term into a temporary instead. EvalMult on a scalar is Clone plus in-place multiply, so this costs one ciphertext copy per nonzero lower-order term (at most k-1, never the top term), fewer copies than guarding in the callers, and it also fixes the two public APIs. Measured at N = 2^15 the extra copies are under 1 ms per function against a 100-300 ms series evaluation.

- take the basis by const reference in internalEvalPolyLinearWithPrecomp and internalEvalChebyshevSeriesLinearWithPrecomp
- drop the now redundant deep copies in EvalFEFuncBootstrapWithPrecomp
- document that EvalPolyWithPrecomp/EvalChebyshevSeriesWithPrecomp leave the precomputation unmodified
- extend UnitTest_MVBReuse to evaluate f1, f2, f1 on one precomputation and add p = 4 rows (FIXEDMANUAL full/sparse, FLEXIBLEAUTO, COMPOSITESCALINGAUTO), which fail without the fix
- add EVAL_PRECOMP_REUSE to UnitTestPoly for the generic power-basis and Chebyshev precomputation APIs


andreea-alexandru (approved):
Looks good. I added some comments to clarify in what conditions the powers can be reused, and saved a clone in FIXEDMANUAL case if levels are not aligned.

